PR #30, opened 2014-11-14
Initial Element models with a basic test

Kapin:
WIP


src/webapp/tests/sana_test_case.py
```diff
@@ -0,0 +1,13 @@
+from django.test import TestCase
+from django.contrib.auth.models import User
+from webapp.models import *
+
+class SanaTestCase(TestCase):
```
Trinovantes: This class should be renamed to be something more specific. Not all tests can assume there's a user/procedure/page

Kapin:
Closing this because I'm going to redo it from scratch.

PR #39, opened 2015-01-11
Setup Grunt

Trinovantes:
- Now automatically compile bootstrap.less when you run 'grunt' in root
  directory. We will still use Django's compressor when generating files for
  production since it automatically rewrites the URL in our templates.


Trinovantes:
This is also a prerequisite for automatically generating Ember templates into a single file (to be implemented in the future)

Kapin:
Update the mac install script and it gets a :+1:

Trinovantes:
I don't have a mac to test `npm install -g grunt-cli`

ConnorCimowsky:
Can confirm that it works on 10.10.2. LGTM!

PR #47, opened 2015-04-17
Renamed folders for organization

Trinovantes:
- Renamed `/sana_builder/` to `/src/`
- Renamed `/sana_builder/sana_builder/` to `/src/sanaprotocolbuilder/`


Trinovantes:
I don't remember if we have auto deploy @ConnorCimowsky . If we do, can you merge this so you can be on call? :laughing:

PR #57, opened 2015-05-12
Decouple Sana logic from homepage

### LINKED ISSUE #55: Pull out editor as separate Django app
The `webapp` app should just manage the static pages while the `editor` app manages the API endpoints.

Trinovantes:
- Renamed `webapp` app to `homepage`
- Created `editor` app

Closes #55


Kapin:
I'm assuming these will share the same database? We need to be able to reference users from procedures for ownership purposes.

Trinovantes:
Same database. You can reference by importing the `User` model from `django.contrib.auth.models`

Trinovantes:
I just noticed I forgot to move `editor.html`. Let me make 1 more commit before reviewing this PR

src/editor/tests/test_user.py
```diff
@@ -0,0 +1,15 @@
+# from django.test import TestCase
```
ConnorCimowsky: Any reason this file is commented out?

ConnorCimowsky:
One question, otherwise LGTM as long as you've tested it locally.

src/editor/tests/test_user.py
Trinovantes: It's going to be moved to a `registration` app that I'm working on. I don't want the merge conflict or a temporary broken test.

Trinovantes: Actually I don't even see why I added this file in the first place...

ConnorCimowsky: Let me know what you decide and I'll re-review. :stuck_out_tongue:

ConnorCimowsky:
LGTM

Kapin:
Same

PR #61, opened 2015-05-21
Add Ember App

ConnorCimowsky:
This is just the default structure as generated by `ember-cli`.


src-frontend/.travis.yml
```diff
@@ -0,0 +1,23 @@
+---
```
Kapin: We should integrate this into our main one later.

Kapin:
:+1:

Trinovantes:
+1

PR #65, opened 2015-05-23
Deleted all frontend related code from backend

### LINKED ISSUE #64: Delete frontend related code from backend

Trinovantes:
- Renamed editor Django app to api
- Deleted static folder in src-backend
- Updated url routing to support api versions
- Removed unnecessary Django apps (flatpages, messages, sessions, sites,
  staticfiles)
- Removed Gruntfile from root directory (since we don't need to compile
  less anymore for the backend)

Closes #64


ConnorCimowsky:
LGTM

PR #81, opened 2015-05-24
Add routes for pages and side-load pages with procedure detail calls

Kapin:


src-backend/api/views.py
```diff
@@ -1,21 +1,39 @@
...
+    def get_queryset(self):
+        self.resource_name = 'procedure'
+        user = self.request.user
+        return models.Procedure.objects.filter(owner=user)
 
-def versionTestView(request, version):
-    return JsonResponse({'version': __parseVersion(version)})
+    def retrieve(self, request, pk=None):
+        queryset = self.get_queryset()
+        self.resource_name = False
```
Trinovantes: add a comment why it's false

ConnorCimowsky:
:+1:

src-backend/api/views.py
```diff
@@ -1,21 +1,39 @@
...
 
-class ProcedureViewSet(viewsets.ModelViewSet):
-    serializer_class = ProcedureSerializer
-    model = Procedure
+        data = {
+            'procedure': serial_proc.data,
+            'pages': page_serial.data
+        }
+
+        return Response(data)
```
Trinovantes: automatically converts to json?

Kapin: yes

src-backend/api/views.py
Kapin: done

Trinovantes:
+1

PR #82, opened 2015-05-24
Enforce positive display_index and fix element ordering

Kapin:


src-backend/api/models.py
```diff
@@ -34,6 +34,8 @@ class Meta:
             ('display_index', 'page')
         )
```
Trinovantes: remove the line break

Trinovantes:
:fries:

ConnorCimowsky:
looks salty to me

PR #83, opened 2015-05-24
Add logging support

Trinovantes:
Now you can log to the console or file with:

```
import logging
logger = logging.getLogger('debugger')
logger.info('Hello World')
```

You can customize the logger in the settings file.


Kapin:
LGTM

PR #85, opened 2015-05-24
Cleaned up settings file

Trinovantes:
Since Django is not serving any pages, we don't need `TEMPLATE_CONTEXT_PROCESSORS `.


Kapin:
:shipit:

PR #86, opened 2015-05-24
Add Elements routing and simplify elements handling

Kapin:
Database wipe is needed from this. Last time we will need to do this.

Redid all the migrations because they broke horribly. Unused fields are hidden on elements.


src-backend/api/serializer.py
```diff
@@ -1,10 +1,47 @@
...
+        fields = [field for field in self.fields.values() if not field.write_only]
+
+        for field in fields:
+            try:
+                attribute = field.get_attribute(instance)
+            except SkipField:
+                continue
+
+            if attribute is not None:
+                ret[field.field_name] = field.to_representation(attribute)
```
Trinovantes: can this the moved into the try block?

Trinovantes:
I don't really like stuffing everything into the `Element` class but if there's no other way, then it's fine. Can we instead use composition pattern?

ConnorCimowsky:
LGTM

Trinovantes:
+1

PR #93, opened 2015-05-25
More frontend stuff

ConnorCimowsky:


src-frontend/app/controllers/add-element-modal.js
```diff
@@ -1,9 +1,34 @@
...
+        eid: this.get('eid'),
+        displayIndex: this.get('displayIndex'),
+        elementType: this.get('elementType'),
+        concept: this.get('concept'),
+        question: this.get('question'),
+        answer: this.get('answer')
+      });
+
+      var choices = this.get('choices');
+      if (choices !== "") {
```
Trinovantes: so we're using a delimited string for the list of choices?

ConnorCimowsky: For the demo, yeah :/ I'll add a TODO

Trinovantes: can use `!!choices` here

src-frontend/app/controllers/procedure.js
```diff
@@ -1,28 +1,48 @@
...
+    addElement: function(page, selectedIndex) {
+      var targetIndex = selectedIndex + 1;
+      var elements = page.get('elements');
+      var elementsToBePushed = elements.filter(function(element) {
+        return element.get('displayIndex') >= targetIndex;
+      });
+
+      elementsToBePushed.forEach(function(element) {
+        element.incrementProperty('displayIndex');
+        element.save();
```
Trinovantes: I guess we can't bulk save here? :(

src-frontend/app/serializers/page.js
```diff
@@ -0,0 +1,13 @@
+import DS from 'ember-data';
+
+export default DS.RESTSerializer.extend({
+  serialize: function(snapshot, options) {
+    var json = this._super(snapshot, options);
+
+    if (json.elements === undefined) {
```
Trinovantes: I think `!json.elements` is better here

Trinovantes:
+1

src-frontend/app/controllers/procedure.js
ConnorCimowsky: I was told that Ember batches them up in the current run loop. But I'm still getting weird behaviour if I cascade a bunch of elements. I'll add a TODO

PR #97, opened 2015-05-30
Move Ember/Django integration into the frontend

ConnorCimowsky:


src-frontend/app/adapters/application.js
```diff
@@ -1,7 +1,6 @@
-import DS from 'ember-data';
+import DRFAdapter from './drf';
```
Trinovantes: what does this do?

ConnorCimowsky: We're now using [this](http://dustinfarris.com/ember-django-adapter/) adapter instead of the default `RESTAdapter`.

So now instead of having Ember-friendly JSON responses from Django with vanilla Ember, it's vanilla Django REST Framework and then this DRF-friendly Ember adapter.

Trinovantes: Sounds good

Trinovantes:
+1

Kapin:
LGTM

PR #98, opened 2015-05-30
Base mockups

Trinovantes:
This is only the base editor. We need to brainstorm and iterate on how to visualize the conditional logic.

screenshot from 2015-05-30 01 06 46


mockups/base.html
```diff
@@ -0,0 +1,36 @@
...
+    <script src="js/less.js"></script>
+    <script src="js/base.js"></script>
+    
+</head>
+<body>
+
+<div class="container">
+    <ul class="procedure">
+        <li>
+            <a href="#" class="add-page">+</a>
```
ConnorCimowsky: `class="add-page octicon octicon-plus"`

ConnorCimowsky:
LGTM

ConnorCimowsky:
"LGTM" —@kapin

PR #104, opened 2015-06-01
Begin implementing first mockup

ConnorCimowsky:


mockups/base.html
```diff
@@ -1,36 +0,0 @@
-<!DOCTYPE html>
```
Trinovantes: why delete this?

src-frontend/app/styles/base.less
```diff
@@ -1,13 +1,13 @@
...
 @import url(constants.less);
-@import url(octicons.less);
 
 //------------------------------------------------------------------------------
 // General
 //------------------------------------------------------------------------------
 
 body{
-    background: #efefef;
+    background: #ececec;
```
Trinovantes: this should be defined as a constant

src-frontend/app/styles/octicons.less
```diff
@@ -1,4 +1,4 @@
-@octicons-font-path: "fonts";
+@octicons-font-path: ".";
```
Trinovantes: fonts should be placed in a separate folder

src-frontend/config/environment.js
```diff
@@ -22,11 +22,11 @@ module.exports = function(environment) {
...
     ENV.APP.API_HOST = 'http://localhost:8000';
     ENV.APP.API_NAMESPACE = 'api';
-    ENV.APP.LOG_RESOLVER = true;
-    ENV.APP.LOG_ACTIVE_GENERATION = true;
-    ENV.APP.LOG_TRANSITIONS = true;
-    ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
-    ENV.APP.LOG_VIEW_LOOKUPS = true;
+    ENV.APP.LOG_RESOLVER = false;
+    ENV.APP.LOG_ACTIVE_GENERATION = false;
+    ENV.APP.LOG_TRANSITIONS = false;
```
Trinovantes: why disable logs in dev?

Trinovantes:
I haven't looked too closely but there seems to be a bit of repeated css.

Also there's no need to merge in a commit that reverts another commit in the same PR

mockups/base.html
ConnorCimowsky: I was trying to delete stuff as I port it over, but I agree I should just leave the mockups until I'm completely done. :sparkles:

src-frontend/app/styles/base.less
ConnorCimowsky: tru

src-frontend/app/styles/octicons.less
ConnorCimowsky: Forgot to move them back. Thanks. :+1:

src-frontend/config/environment.js
ConnorCimowsky: They print hundreds of lines for a simple page load, and don't provide any useful information. Not worth having them enabled all the time, IMO

ConnorCimowsky:
I had done the merge/revert so that I would have that work recorded somewhere, but it looks like we're not going to take that approach anyway. Will rebase.

ConnorCimowsky:
Re: CSS, maybe you can help me squash some of it later today. :innocent:

src-frontend/config/environment.js
Trinovantes: can you assign these with a local var (e.g. `var enableLogs = false`) so it'll be easier to toggle later on

Trinovantes:
@Kapin what do you think about merging in revert commits? Also now I see how you got so many commits in master :p

Kapin:
Avoid revert commits unless it's a specific revert. Within a PR do an interactive rebase :smile:

Issue #102, opened 2015-06-01
Ensure all APIs have end-to-end tests


Kapin:
Should have tests that verify that hitting the API works. These can either live in the front-end, back-end or be a separate model as acceptance tests

Trinovantes:
I think this is finished?

Trinovantes:
I think all of our current API endpoints have integration tests now? @Kapin

PR #105, opened 2015-06-01
Add unit tests for procedures and pages

Kapin:


src-backend/registration/tests/test_user.py
```diff
@@ -10,8 +10,5 @@ def setUp(self):
         self.test_user.save()
 
     def test_user_has_token(self):
-        try:
-            token = Token.objects.get(user=self.test_user)
-            self.assertFalse(token is None)
-        except ObjectDoesNotExist:
-            self.assertTrue(False)
+        token = Token.objects.get(user=self.test_user)
```
Trinovantes: won't this throw an exception?

Trinovantes:
Should also create a test for the output json if possible

src-backend/registration/tests/test_user.py
Kapin: if it does, it'll fail.

Kapin:
@Trinovantes that is a functional test vs. unit test. See #102

Trinovantes:
This should be named functional tests to be more specific

src-backend/registration/tests/test_user.py
Ommy: Do exceptions cause the tests to fail in a functional sense, or does the test crash? I feel like it would be better to catch exceptions

Kapin: http://stackoverflow.com/questions/7799593/how-an-exceptions-be-tested-with-nose

Ommy:
Looks fine, +1

Trinovantes:
+1

ConnorCimowsky:
yes-approval-thumbs-up-brent-rambo-1

PR #116, opened 2015-06-05
Add page reordering

ConnorCimowsky:


src-frontend/app/components/page-list.js
```diff
@@ -0,0 +1,35 @@
...
+    didInsertElement: function() {
+        var self = this;
+        this.$("ul.procedure").sortable({
+            items: 'li:not(:first-child)',
+            handle: 'span.drag-handle',
+            revert: true,
+            scroll: true,
+            placeholder: 'ui-state-highlight',
+            update: function(event, ui) {
+                var indices = {};
```
Trinovantes: Shouldn't this be an array?

src-frontend/app/components/page-list.js
```diff
@@ -0,0 +1,35 @@
+export default Ember.Component.extend({
+    didInsertElement: function() {
+        var self = this;
```
Trinovantes: should just call this emberComponent

Kapin:
Non-CSS/LESS makes sense. lgtm

Trinovantes:
+1

Ommy:
+1 (y)

Issue #48, opened 2015-04-18
Use Python's XML ETree Library to generate protocols


Kapin:
See https://docs.python.org/2/library/xml.etree.elementtree.html#building-xml-documents.

This should make things wayyyyy easier.

Trinovantes:
Resolved by #134

PR #120, opened 2015-06-06
Create the protocol generator

Kapin:
Generator that uses the database to create xml string output. Includes a lot of tests. I will create the routes, and other bits later.

Tests need a lot of future cleanup as there is a ridiculous amount of redundant code


src-backend/api/generator.py
```diff
@@ -0,0 +1,97 @@
...
+        props = self._get_properties()
+        return ElementTree.SubElement(parent, self.name, props)
+
+
+class ProtocolBuilder:
+    @classmethod
+    def _prettify(cls, elem):
+        rough_string = ElementTree.tostring(elem, 'utf-8')
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent="  ")
```
Trinovantes: 4 spaces

src-backend/api/generator.py
```diff
@@ -0,0 +1,97 @@
...
+    def _generateTree(cls, owner, pk):
+        try:
+            procedure = Procedure.objects.get(pk=pk)
+        except Procedure.DoesNotExist:
+            raise ValueError('Invalid procedure id')
+
+        if procedure.owner != owner:
+            raise ValueError('Invalid owner')
+
+        procedureElement = ProcedureGenerator(procedure).generate()
```
Trinovantes: snake case

src-backend/api/tests/test_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        test_user.save()
+
+        self.procedure = Procedure.objects.create(
+            author='tester',
+            title='test procedure',
+            owner=test_user
+        )
+
+        self.generator = ProcedureGenerator(self.procedure)
+        self.procedureElement = self.generator.generate()
```
Trinovantes: snake case

src-backend/api/generator.py
```diff
@@ -0,0 +1,97 @@
...
+        return props
+
+    def generate(self, parent):
+        props = self._get_properties()
+        return ElementTree.SubElement(parent, self.name, props)
+
+
+class ProtocolBuilder:
+    @classmethod
+    def _prettify(cls, elem):
```
Trinovantes: if these are supposed to be private, should be prefixed with `__`

src-backend/api/tests/test_generator.py
```diff
@@ -0,0 +1,274 @@
...
+            concept='COMPLAINT PICTURE',
+            question='Add a Picture:',
+            answer='',
+            page=page4
+        )
+
+        self.procedure.save()
+
+    def test_generates_tree(self):
+        tree = ProtocolBuilder._generateTree(self.test_user, self.procedure.id)
```
Trinovantes: calling private method?

Trinovantes:
Resolves #48

src-backend/api/generator.py
Kapin: will fix

src-backend/api/generator.py
Kapin: http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python

Seems to indicate this is the correct convention.

src-backend/api/tests/test_generator.py
Kapin: It's a test. There is no concept of private in Python, it's only a thing saying "don't touch this".

src-backend/api/generator.py
Trinovantes: I don't see why we shouldn't use `__` to be strict. If this is supposed to be conceptually private, it shouldn't be easily called by name.

src-backend/api/tests/test_generator.py
Trinovantes: I think this method should either be public or we only test the pre/post conditions of any public methods that will call this internally (http://c2.com/cgi/wiki?InappropriateIntimacy)

Kapin: I have no issues making it public. Will change

src-backend/api/generator.py
Kapin: Fair point.

Trinovantes:
+1

Trinovantes:
Does this include a Django view endpoint that responds with the xml?

Kapin:
@Trinovantes it does not. I have a "prototype" working on my machine, but it's not serializing the xml properly.

Kapin:
Closing in favour of #134

PR #124, opened 2015-06-06
Add timestamps for created and last_modified

Kapin:


src-backend/api/serializer.py
```diff
@@ -17,7 +17,9 @@ class Meta:
             'page',
             'choices',
             'numeric',
-            'element_type'
+            'element_type',
+            'last_modified',
```
Ommy: maybe these should be worded `last_modified_at` and `created_at` to follow how Rails does it

Trinovantes: I think it's a python convent to use `created` and `last_modified`

src-backend/api/serializer.py
```diff
@@ -48,7 +50,9 @@ class Meta:
             'id',
             'display_index',
             'procedure',
-            'elements'
+            'elements',
+            'last_modified',
+            'created'
```
Trinovantes: we should have a common convention to have a comma on the last element of a list (in both JS and Python) so we don't have -1 +1 commits

Trinovantes: do this in a separate PR

Trinovantes:
+1

Ommy:
+1

ConnorCimowsky:
LGTM

PR #128, opened 2015-06-07
Modified install docs to account for bundler

Ommy:
Currently doing bundle install inside chef requires 1.9.3 Ruby.
Also, bundler is a gem, so we need to get gem and then do a `gem install bundler`

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -9,7 +9,7 @@ We use Chef to configure our servers and automate our deployments.
 On your local machine, you will need the following packages to configure a server with Chef:
 
 - Git
-- Ruby
+- Ruby (>= 1.9.3)
 - Bundler
 - Python
 - Fabric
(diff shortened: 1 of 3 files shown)


Trinovantes:
We should just break up `INSTALL.md` into two files:
- `SETUP_DEVELOPMENT.md`
- `DEPLOY.md`

ConnorCimowsky:
LGTM

docs/INSTALL.md
```diff
@@ -18,7 +18,8 @@ This will vary depending on your OS, but here's an example using Ubuntu:
 
 ```shell
 apt-get update
-apt-get install git bundler python-pip
+apt-get install git gem python-pip
```
Kapin: Doesn't ruby come with RubyGems?

Kapin:
One question otherwise :+1:

docs/INSTALL.md
Ommy: Ah you're right, Ruby 1.9+ comes with RubyGems installed. I'll remove it.

Trinovantes:
+1

PR #129, opened 2015-06-07
Disabled CORS for development on Ember server

### LINKED ISSUE #100: Stop using CORS for frontend

Trinovantes:
Resolves #100

### DIFF
--- a/src-frontend/config/environment.js
+++ b/src-frontend/config/environment.js
@@ -19,16 +19,27 @@ module.exports = function(environment) {
     }
   };
 
-  var enableLogging = false;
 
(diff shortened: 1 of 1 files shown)


Ommy:
+1

ConnorCimowsky:
Can you explain what this is doing? Don't we need CORS for local development?

Trinovantes:
This gets rid of the CORS errors that Ember keeps printing in the console during development since we're making a request from `localhost:4200` to `localhost:8000`. 

Normally, we want to enable it in production for security purposes. But that's something that nginx would do instead of Ember CLI anyways.

PR #131, opened 2015-06-07
Add middleware to print 500 errors to console

Trinovantes:
This should help with debugging errors when we're making AJAX requests to the server


ConnorCimowsky:
LGTM

Kapin:
Can we please enable this only with debug mode? I'd prefer to not have us dumping stack traces to the console that could have database object output in them on a production server.

Trinovantes:
Updated

PR #133, opened 2015-06-19
Update indices of Pages on backend instead of batch puts on the frontend

### LINKED ISSUE #108: Investigate if we can send a single POST request to insert page
When inserting a new page at index `i`, we are currently sending multiple POST requests to update the indices of pages `i` to `n`. Ideally we should just send a simple POST request to insert that page and let the server handle reordering the page indices.

Right now, we're relaxing the database's unique constraint on `display_index` to reorder pages because Ember sends a couple of async POST requests to update the indices and we can sometimes get out of order updates.

Ommy:
resolves #108

### DIFF
--- a/src-backend/api/signals/page_pre_save.py
+++ b/src-backend/api/signals/page_pre_save.py
@@ -1,0 +1,16 @@
+from django.dispatch import receiver
+from django.db.models.signals import pre_save
+from django.db.models import F
+from api.models import Page
+import logging
+
+logger = logging.getLogger('debugger')
+
+
+@receiver(pre_save, sender=Page)
+def on_page_pre_save(sender, instance=None, raw=False, **kwargs):
+    # Update index of all objects with display index >= object we are saving
+    if instance.display_index is not None:
+        logger.info("Updating display indices >= {0}".format(instance.display_index))
+        Page.objects.filter(display_index__gte=instance.display_index)\
+                    .update(display_index=(F('display_index') + 1))
--- a/src-backend/api/signals/page_pre_save.py
+++ b/src-backend/api/signals/page_pre_save.py
@@ -4,6 +4,7 @@ from django.db.models import F
 from api.models import Page
 import logging
 
+
 logger = logging.getLogger('debugger')
 
 
@@ -11,6 +12,7 @@ logger = logging.getLogger('debugger')
 def on_page_pre_save(sender, instance=None, raw=False, **kwargs):
     # Update index of all objects with display index >= object we are saving
     if instance.display_index is not None:
-        logger.info("Updating display indices >= {0}".format(instance.display_index))
+        logger.info("Updating display indices >= {index} for Page {id}"
+                    .format(index=instance.display_index, id=instance.id))
         Page.objects.filter(display_index__gte=instance.display_index)\
                     .update(display_index=(F('display_index') + 1))
--- a/src-backend/api/tests/test_model_page.py
+++ b/src-backend/api/tests/test_model_page.py
@@ -3,6 +3,7 @@ from django.contrib.auth.models import User
 from django.db import IntegrityError
 from nose.tools import raises, assert_equals, assert_not_equals, nottest
 from api.models import Procedure, Page
+from api.signals import page_pre_save  # noqa
 
 
 class ProcedureTest(TestCase):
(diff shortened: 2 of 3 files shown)


src-backend/api/urls.py
```diff
@@ -1,6 +1,7 @@
 from django.conf.urls import url, include
 from views import ProcedureViewSet, PageViewSet, ElementViewSet
 from rest_framework import routers
+from signals import page_pre_save
```
Ommy: The reason why I added this here is because we have to import the `page_pre_save` file somewhere, but we can't do it in `models.py` since the `page_pre_save.py` file requires it, which would result in circular dependency. So we load it here in `urls.py`

src-backend/api/signals/page_pre_save.py
```diff
@@ -0,0 +1,16 @@
+from django.dispatch import receiver
+from django.db.models.signals import pre_save
+from django.db.models import F
+from api.models import Page
+import logging
+
```
Trinovantes: Shouldn't this be 2 lines from the imports? Or is that the linter rule?

src-backend/api/signals/page_pre_save.py
```diff
@@ -0,0 +1,16 @@
...
+logger = logging.getLogger('debugger')
+
+
+@receiver(pre_save, sender=Page)
+def on_page_pre_save(sender, instance=None, raw=False, **kwargs):
+    # Update index of all objects with display index >= object we are saving
+    if instance.display_index is not None:
+        logger.info("Updating display indices >= {0}".format(instance.display_index))
```
Trinovantes: Should also log the page id

src-backend/api/signals/page_pre_save.py
```diff
@@ -0,0 +1,16 @@
...
+from api.models import Page
+import logging
+
+logger = logging.getLogger('debugger')
+
+
+@receiver(pre_save, sender=Page)
+def on_page_pre_save(sender, instance=None, raw=False, **kwargs):
+    # Update index of all objects with display index >= object we are saving
+    if instance.display_index is not None:
```
Trinovantes: Does this mean if we try to save a page with a `display_index` it'll do the updates?

src-backend/api/tests/test_model_page.py
```diff
@@ -88,3 +89,20 @@ def test_page_ordering(self):
 
         for i in range(0, 4):
             assert_equals(pages[i], expected_pages[i])
+
+    def test_page_reordering(self):
```
Trinovantes: Add a couple more test cases such as inserting before/after existing ids

Trinovantes:
Should squash this PR since you have 1 feature commit and 3 "fixes". Also fix your PR title.

src-backend/api/tests/test_model_page.py
```diff
@@ -88,3 +89,20 @@ def test_page_ordering(self):
...
+        page1 = Page.objects.create(
+            display_index=0,
+            procedure=self.test_procedure1
+        )
+
+        page2 = Page.objects.create(
+            display_index=0,
```
ConnorCimowsky: Maybe give this one a `display_index` of 1? Would be more realistic since we'll have integrity constraints anyway.

ConnorCimowsky:
LGTM after Stephen's comments!

src-backend/api/signals/page_pre_save.py
Ommy: So are we sending back every attribute in a `PUT` everytime? If that's the case, then I can add extra checks to see if that attribute is actually being modified or not.

src-backend/api/signals/page_pre_save.py
Ommy: Linter didn't complain, but if you really want I can add it

Trinovantes:
+1

src-backend/api/signals/page_pre_save.py
Kapin: @ConnorCimowsky would know this

src-backend/api/tests/test_model_page.py
Kapin: That's what this test is doing. Existing tests cover inserting after existing.

src-backend/api/signals/page_pre_save.py
```diff
@@ -0,0 +1,18 @@
+from django.dispatch import receiver
+from django.db.models.signals import pre_save
+from django.db.models import F
+from api.models import Page
+import logging
+
+
+logger = logging.getLogger('debugger')
```
Kapin: only main question is that does this mean it runs in debug mode only?

src-backend/api/signals/page_pre_save.py
```diff
@@ -0,0 +1,18 @@
+from django.dispatch import receiver
+from django.db.models.signals import pre_save
+from django.db.models import F
```
Kapin: I'm probably over tired, but :laughing:

src-backend/api/signals/page_pre_save.py
Ommy: According to https://docs.djangoproject.com/en/1.8/topics/logging/#django-s-default-logging-configuration it seems like these logs will only be logged on debug mode, but @Trinovantes can confirm for sure.

Kapin:
LGTM

PR #134, opened 2015-06-19
Generator API and file saving

Kapin:
Javascript part is a tad hacky, but that is the recommended way of doing it.

Still needs styling work.

Depends on #120

### DIFF
--- a/src-backend/api/generator.py
+++ b/src-backend/api/generator.py
@@ -1,6 +1,6 @@
 from xml.etree import ElementTree
 from xml.dom import minidom
-from models import *
+from models import Procedure
 
 
 class ProcedureGenerator:
@@ -8,7 +8,7 @@ class ProcedureGenerator:
         self.name = 'Procedure'
         self.procedure = procedure
 
-    def _get_properties(self):
+    def __get_properties(self):
         props = {
             'title': self.procedure.title,
             'author': self.procedure.author
@@ -24,7 +24,7 @@ class ProcedureGenerator:
 
     def generate(self):
         element = ElementTree.Element(self.name)
-        element.attrib = self._get_properties()
+        element.attrib = self.__get_properties()
 
         return element
 
@@ -43,7 +43,7 @@ class ElementGenerator:
         self.name = 'Name'
         self.element = element
 
-    def _get_properties(self):
+    def __get_properties(self):
         props = {
             'type': self.element.element_type,
             'id': self.element.eid,
@@ -61,19 +61,19 @@ class ElementGenerator:
         return props
 
     def generate(self, parent):
-        props = self._get_properties()
+        props = self.__get_properties()
         return ElementTree.SubElement(parent, self.name, props)
 
 
 class ProtocolBuilder:
     @classmethod
-    def _prettify(cls, elem):
+    def __prettify(cls, elem):
         rough_string = ElementTree.tostring(elem, 'utf-8')
         reparsed = minidom.parseString(rough_string)
-        return reparsed.toprettyxml(indent="  ")
+        return reparsed.toprettyxml(indent="    ")
 
     @classmethod
-    def generate(cls, owner, pk):
+    def generateETree(cls, owner, pk):
         try:
             procedure = Procedure.objects.get(pk=pk)
         except Procedure.DoesNotExist:
@@ -82,12 +82,16 @@ class ProtocolBuilder:
         if procedure.owner != owner:
             raise ValueError('Invalid owner')
 
-        procedureElement = ProcedureGenerator(procedure).generate()
+        procedure_element = ProcedureGenerator(procedure).generate()
 
-        for page in procedure.pages:
-            pageElement = PageGenerator(page).generate(procedureElement)
+        for page in procedure.pages.all():
+            page_element = PageGenerator(page).generate(procedure_element)
 
-            for element in page.elements:
-                ElementGenerator(element).generate(pageElement)
+            for element in page.elements.all():
+                ElementGenerator(element).generate(page_element)
 
-        return cls._prettify(procedureElement)
+        return procedure_element
+
+    @classmethod
+    def generate(cls, owner, pk):
+        return cls.__prettify(cls.generateETree(owner, pk))
--- a/src-backend/api/generator.py
+++ b/src-backend/api/generator.py
@@ -62,7 +62,7 @@ class ElementGenerator:
 
     def generate(self, parent):
         props = self._get_properties()
-        ElementTree.SubElement(parent, self.name, props)
+        return ElementTree.SubElement(parent, self.name, props)
 
 
 class ProtocolBuilder:
--- a/src-backend/api/generator.py
+++ b/src-backend/api/generator.py
@@ -1,0 +1,93 @@
+from xml.etree import ElementTree
+from xml.dom import minidom
+from models import *
+
+
+class ProcedureGenerator:
+    def __init__(self, procedure):
+        self.name = 'Procedure'
+        self.procedure = procedure
+
+    def _get_properties(self):
+        props = {
+            'title': self.procedure.title,
+            'author': self.procedure.author
+        }
+
+        if not procedure.version:
+            props['version'] = self.procedure.version
+
+        if not procedure.uuid:
+            props['uuid'] = self.procedure.uuid
+
+        return props
+
+    def generate(self):
+        element = ElementTree.Element(self.name)
+        element.attrib = self._get_properties()
+
+        return element.attrib
+
+
+class PageGenerator:
+    def __init__(self, page):
+        self.name = 'Page'
+        self.page = page
+
+    def generate(self, parent):
+        return ElementTree.SubElement(parent, self.name)
+
+
+class ElementGenerator:
+    def __init__(self, element):
+        self.name = 'Name'
+        self.element = element
+
+    def _get_properties(self):
+        props = {
+            'type': element.element_type,
+            'id': element.eid,
+            'concept': element.concept,
+            'question': element.question,
+            'answer': element.answer
+        }
+
+        if not element.numeric:
+            props['numeric'] = self.element.numeric
+
+        if not element.choices:
+            props['choices'] = self.element.choices
+
+        return props
+
+    def generate(self, parent):
+        props = self._get_properties()
+        ElementTree.SubElement(parent, self.name, props)
+
+
+class ProtocolBuilder:
+    @classmethod
+    def _prettify(cls, elem):
+        rough_string = ElementTree.tostring(elem, 'utf-8')
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent="  ")
+
+    @classmethod
+    def generate(cls, owner, pk):
+        try:
+            procedure = Procedure.objects.get(pk=pk)
+        except Procedure.DoesNotExist:
+            raise ValueError('Invalid procedure id')
+
+        if procedure.owner != owner:
+            raise ValueError('Invalid owner')
+
+        procedureElement = ProcedureGenerator(procedure).generate()
+
+        for page in procedure.pages:
+            pageElement = PageGenerator(page).generate(procedureElement)
+
+            for element in page.elements:
+                ElementGenerator(element).generate(pageElement)
+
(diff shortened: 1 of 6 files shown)


src-backend/api/views.py
```diff
@@ -13,6 +17,18 @@ def get_queryset(self):
         user = self.request.user
         return models.Procedure.objects.filter(owner=user)
 
+    @detail_route(methods=['get'])
+    def generate(self, request, pk=None):
+        try:
+            protocol = ProtocolBuilder.generate(request.user, pk)
+        except ValueError:
+            return Response('Invalid procedure', status=status.HTTP_404_NOT_FOUND)
```
Trinovantes: Should use this instead https://docs.djangoproject.com/en/1.8/topics/http/views/#the-http404-exception

src-frontend/app/controllers/procedure.js
```diff
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import ENV from 'src-frontend/config/environment';
```
Trinovantes: @ConnorCimowsky is this the correct path for importing `ENV`?

src-frontend/app/controllers/procedure.js
```diff
@@ -63,6 +64,29 @@ export default Ember.ObjectController.extend({
             });
 
             this.endPropertyChanges();
+        },
+        generateProtocol: function() {
+            var filename = this.get('model.title') + '.xml';
+            $.ajax({
+                beforeSend: function(xhr) {
+                    xhr.setRequestHeader('Authorization', 'Token ' + Cookies.get('authorizationToken'));
```
Trinovantes: I thought @Ommy configured all of our AJAX calls to automatically insert the token? If not, we should have our own custom AJAX caller that automatically sets our default config.

src-frontend/app/controllers/procedure.js
```diff
@@ -63,6 +64,29 @@ export default Ember.ObjectController.extend({
...
+                },
+                type: 'GET',
+                url: ENV.APP.API_HOST + '/' + ENV.APP.API_NAMESPACE + '/procedures/' + this.get('model.id') + '/generate/',
+                success: function(data) {
+                    var blob = new Blob([(new XMLSerializer()).serializeToString(data)], { type: 'text/xml' });
```
Trinovantes: Still very hacky but make sure you create an issue to fix it later and make a comment about it here.

src-backend/api/generator.py
```diff
@@ -0,0 +1,97 @@
...
+        props = self.__get_properties()
+        return ElementTree.SubElement(parent, self.name, props)
+
+
+class ProtocolBuilder:
+    @classmethod
+    def __prettify(cls, elem):
+        rough_string = ElementTree.tostring(elem, 'utf-8')
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent="    ")
```
Trinovantes: Isn't there a python function to repeat a character n times? We can define indentation size as a parameter in user settings later.

src-backend/api/tests/test_generator.py
```diff
@@ -0,0 +1,274 @@
...
+            eid='1',
+            element_type='SELECT',
+            concept='HEART SURGERY',
+            question='Which valve',
+            answer='',
+            page=page
+        )
+
+        self.generator = ElementGenerator(self.element)
+        self.elementElement = self.generator.generate(ElementTree.Element('test'))
```
Trinovantes: what?

src-backend/api/tests/test_generator.py
```diff
@@ -0,0 +1,274 @@
...
+
+    @raises(ValueError)
+    def test_invalid_owner(self):
+        bad_user = User.objects.create_user(
+            'A Bad user',
+            'bad@sanaprotocolbuilder.me',
+            'bad'
+        )
+
+        ProtocolBuilder.generate(bad_user, self.procedure.id)
```
Trinovantes: We should also have a test to ensure this generates a valid XML file.

src-backend/api/views.py
Kapin: agreed

src-frontend/app/controllers/procedure.js
Kapin: Wasn't working for me. Can investigate further

src-frontend/app/controllers/procedure.js
Kapin: I'm going to refactor this a bit so it looks less gross

src-backend/api/tests/test_generator.py
Kapin: they are all element tree elements. It's the naming convention I chose. I can rename it

src-backend/api/tests/test_generator.py
Kapin: that's harder to do and requires a parser to ensure it works properly

src-backend/api/tests/test_generator.py
Trinovantes: maybe `EtreeElement`

src-frontend/app/controllers/procedure.js
Ommy: That's the path to the `environment.js` file so I'm going to say yes.

src-frontend/app/controllers/procedure.js
Ommy: I didn't do that, wasn't in the scope of any tasks I've done. My requests for things like signup when through cause we were hitting `/auth/` which doesn't require a token.

src-backend/api/generator.py
Ommy: `"{character}"*{n times}"` works.

src-frontend/app/templates/procedure.hbs
```diff
@@ -1,3 +1,4 @@
 <div class="container-fluid">
+    <a class="add-page" {{action "generateProtocol"}}>Generate</a>
```
Ommy: So we're no longer doing a button?

src-frontend/app/controllers/procedure.js
```diff
@@ -63,6 +64,29 @@ export default Ember.ObjectController.extend({
...
             this.endPropertyChanges();
+        },
+        generateProtocol: function() {
+            var filename = this.get('model.title') + '.xml';
+            $.ajax({
+                beforeSend: function(xhr) {
+                    xhr.setRequestHeader('Authorization', 'Token ' + Cookies.get('authorizationToken'));
+                },
+                type: 'GET',
+                url: ENV.APP.API_HOST + '/' + ENV.APP.API_NAMESPACE + '/procedures/' + this.get('model.id') + '/generate/',
```
Ommy: Can we get this to be a bit cleaner? Maybe add some new entries into `ENV`?

src-frontend/app/controllers/procedure.js
```diff
@@ -63,6 +64,29 @@ export default Ember.ObjectController.extend({
...
+                success: function(data) {
+                    var blob = new Blob([(new XMLSerializer()).serializeToString(data)], { type: 'text/xml' });
+                    var url = window.URL.createObjectURL(blob);
+                    var a = document.createElement("a");
+                    var randomId = Math.random().toString();
```
Ommy: Why do we need to use a random number string for an ID vs simply giving it an explicit ID?

src-frontend/app/controllers/procedure.js
Trinovantes: Alright then it's fine. I was concerned that it's using absolute paths

ConnorCimowsky: Yup looks good

src-frontend/app/controllers/procedure.js
ConnorCimowsky: Yeah I think we were going to hold off on that until we get a chance to evaluate [ember-simple-auth](https://github.com/simplabs/ember-simple-auth)

src-frontend/app/controllers/procedure.js
ConnorCimowsky: I will add a chore to fix this in my own code as well, but I think you can/should be using `Ember.$` instead of `$`.

src-frontend/app/templates/procedure.hbs
ConnorCimowsky: If you just add `style="btn btn-default"` it will be styled by Bootstrap as a button

src-frontend/app/controllers/procedure.js
ConnorCimowsky: Yeah, I agree we should add a `ENV.APP.GENERATOR_URL` or something

Kapin: Agreed as well.

src-frontend/app/templates/procedure.hbs
Kapin: :+1:

src-backend/api/generator.py
Kapin: good to know. Will change

src-frontend/app/controllers/procedure.js
Kapin: I can change that.

src-frontend/app/controllers/procedure.js
Kapin: I was debating between that and an explicit id. I was worried about deleting things that we didn't want to delete and some weird dom issues. I agree that is much simpler.

Ommy:
Maybe you could squash some of those commits. Other then that, looks good to me

src-backend/api/tests/test_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+
+        Element.objects.create(
+            display_index=0,
+            eid='burns',
+            element_type='MULTI_SELECT',
+            concept='BURNS',
+            question='Select One or More of the Following:',
+            answer='',
+            choices='On the head,On the chest,Loins,Inhalation burns,Chemical,In an enclosure,Patient Unconscious,Epileptic Patient, Diabetic Patient',  # noqa
```
ConnorCimowsky: Sorry for the random question, in what may be the wrong place: but aren't we going to be storing `choices` as a JSON-formatted array?

src-frontend/app/controllers/procedure.js
```diff
@@ -63,6 +64,31 @@ export default Ember.ObjectController.extend({
...
+                success: function(data) {
+                    var blob = new Blob([(new XMLSerializer()).serializeToString(data)], { type: 'text/xml' });
+                    var url = window.URL.createObjectURL(blob);
+
+                    var a = document.createElement("a");
+                    document.body.appendChild(a);
+                    a.setAttribute('id', 'generate-trigger');
+                    a.href = url;
+                    a.download = filename;
+                    a.click();
```
ConnorCimowsky: What's this code doing? Maybe you can explain in person?

ConnorCimowsky:
LGTM after comments

src-backend/api/tests/test_generator.py
Trinovantes: they should (or *need*) to be stored in a format that can serialize an arbitrary string array (i.e. comma delimited probably wont' be enough)

Kapin: I need to make changes to the element serializer to handle the choices properly. Let's talk about this today.

src-frontend/app/controllers/procedure.js
Kapin: It's a "hack" to allow us to save a file from an ajax call. If we can find a way to navigate to api page and have it behave properly, we don't need this.

Kapin:
@Trinovantes?

src-backend/api/generator.py
```diff
@@ -0,0 +1,97 @@
...
+
+class ProtocolBuilder:
+    @classmethod
+    def __prettify(cls, elem):
+        rough_string = ElementTree.tostring(elem, 'utf-8')
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent=' ' * 4)
+
+    @classmethod
+    def generateETree(cls, owner, pk):
```
Trinovantes: snake case

Trinovantes:
+1

PR #140, opened 2015-06-22
Clean up page indexing

ConnorCimowsky:
Now that display indices are updated in the backend, we can get rid of the old client code.

Note: reordering is still broken. this is just for adding new pages

### DIFF
--- a/src-frontend/app/controllers/edit-page-modal.js
+++ b/src-frontend/app/controllers/edit-page-modal.js
@@ -12,7 +12,7 @@ export default Ember.Controller.extend({
 
             var newElement = this.store.createRecord('element', {
                 page: page,
-                displayIndex: targetIndex
+                displayIndex: selectedIndex
             });
 
             newElement.save().then(function() {
--- a/src-frontend/app/controllers/edit-page-modal.js
+++ b/src-frontend/app/controllers/edit-page-modal.js
@@ -8,27 +8,23 @@ export default Ember.Controller.extend({
 
     actions: {
         addElement: function(selectedIndex) {
-            var targetIndex = selectedIndex + 1;
-            var elements = this.get('model.elements');
-            var elementsToBePushed = elements.filter(function(element) {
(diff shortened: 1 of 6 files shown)


Kapin:
LGTM

Ommy:
+1

src-frontend/app/controllers/edit-page-modal.js
```diff
@@ -8,27 +8,23 @@ export default Ember.Controller.extend({
...
-                element.incrementProperty('displayIndex');
-                // TODO: Find a way to batch these updates
-                element.save();
-            });
+            var page = this.get('model');
 
             var newElement = this.store.createRecord('element', {
-                page: this.get('model'),
+                page: page,
                 displayIndex: targetIndex
```
Trinovantes: You deleted `targetIndex`. Shouldn't this be `selectedIndex`?

src-frontend/tests/unit/helpers/plus-one-test.js
```diff
@@ -0,0 +1,10 @@
+import { plusOne } from '../../../helpers/plus-one';
+import { module, test } from 'qunit';
+
+module('Unit | Helper | plus one');
+
+// Replace this with your real tests.
+test('it works', function(assert) {
+  var result = plusOne(42);
+  assert.ok(result);
```
Trinovantes: shouldn't this test the value of `result`?

src-frontend/app/controllers/edit-page-modal.js
ConnorCimowsky: Yup. Haven't really fixed the element stuff yet, but this is definitely wrong. Thanks!

src-frontend/tests/unit/helpers/plus-one-test.js
ConnorCimowsky: Eventually. This is just boilerplate from `ember generate`

PR #141, opened 2015-06-23
Add a custom field to serialize choices properly

Kapin:
Fixes all the issues with storing choices in the database and retrieving
them

### DIFF
--- a/src-backend/api/generator.py
+++ b/src-backend/api/generator.py
@@ -1,6 +1,7 @@
 from xml.etree import ElementTree
 from xml.dom import minidom
 from models import Procedure
+import json
 
 
 class ProcedureGenerator:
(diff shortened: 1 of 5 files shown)


src-backend/api/generator.py
```diff
@@ -56,10 +57,14 @@ def __get_properties(self):
             props['numeric'] = self.element.numeric
 
         if self.element.choices:
-            props['choices'] = self.element.choices
+            props['choices'] = self.__parse_choices()
 
         return props
 
+    def __parse_choices(self):
+        choice_list = json.loads(self.element.choices)
```
Trinovantes: so our database stores the choices as a json array?

src-backend/api/generator.py
```diff
@@ -56,10 +57,14 @@ def __get_properties(self):
...
+        return ','.join(choice_list)
```
Trinovantes: what does this return (in terms of format)?

src-backend/api/generator.py
Kapin: currently yes, this is probably going to change.

src-backend/api/generator.py
Kapin: choice1,choice2,choice3

Trinovantes:
+1

ConnorCimowsky:
LGTM

PR #144, opened 2015-06-27
Use simple-ember-auth for more robust authentication and session management

ConnorCimowsky:
Also, rip out js-cookie

bugs-bunny-florida


ConnorCimowsky:
@Ommy, you should probably look at this since you did the login stuff :innocent:

src-frontend/app/authenticators/token.js
```diff
@@ -0,0 +1,37 @@
...
+                method: 'POST',
+                url: ENV.APP.API_HOST + '/auth/login',
+                data: credentials
+            }).then(function(response) {
+                Ember.run(function() {
+                    resolve({ token: response.token });
+                });
+            }, function(xhr, status, error) {
+                Ember.run(function() {
+                    reject(xhr);
```
Ommy: What does the reject function do?

src-frontend/app/authenticators/token.js
```diff
@@ -0,0 +1,37 @@
...
+                reject();
+            }
+        });
+    },
+
+    authenticate(credentials) {
+        return new Ember.RSVP.Promise(function(resolve, reject) {
+            Ember.$.ajax({
+                method: 'POST',
+                url: ENV.APP.API_HOST + '/auth/login',
```
Trinovantes: maybe the login url should be a constant in the ENV?

src-frontend/app/authorizers/token.js
```diff
@@ -0,0 +1,14 @@
+import Base from 'simple-auth/authorizers/base';
+
+export default Base.extend({
+    authorize: function(jqXHR, requestOptions) {
+        var session = this.get('session');
+
+        if (session.isAuthenticated) {
```
Trinovantes: how does it know it's authenticated?

src-frontend/bower.json
```diff
@@ -11,9 +11,9 @@
     "ember-resolver": "~0.1.15",
     "jquery": "^1.11.1",
     "jquery-ui": "~1.11.4",
-    "js-cookie": "^1.5.1",
     "loader.js": "ember-cli/loader.js#3.2.0",
     "qunit": "~1.17.1",
-    "bootstrap": "~3.3.4"
+    "bootstrap": "~3.3.4",
+    "ember-simple-auth": "0.8.0"
```
Trinovantes: we should end this line with a comma so if we need to add more libraries, it won't show up as a deletion and insertion

Ommy:
LGTM after questions

Kapin:
LGTM once comments from others are addressed.

src-frontend/app/authenticators/token.js
ConnorCimowsky: This causes the promise to go unfulfilled, which tells `ember-simple-auth` that authentication was unsuccessful. We don't currently do anything in this case, but we already have a TODO for showing a banner or something.

src-frontend/app/authenticators/token.js
ConnorCimowsky: Agreed

src-frontend/app/authorizers/token.js
ConnorCimowsky: A successful authentication in `app/authenticators/token.js` will cause this to be set to `true` automatically :)

src-frontend/bower.json
ConnorCimowsky: :ok:

PR #148, opened 2015-06-28
Make everything nullable and blankable, fix a couple of things in the frontend

ConnorCimowsky:
This was the stuff we did leading up to the demo last week.


Trinovantes:
Will also need to extend the generator to validate the possibly empty fields

src-frontend/app/templates/components/page-list.hbs
```diff
@@ -6,7 +6,7 @@
                 <span class="drag-handle octicon octicon-three-bars"></span>
                 <span class="page-label">Page {{page.id}}, Display {{page.displayIndex}}</span>
                 <span class="add-condition mega-octicon octicon-diff-renamed"></span>
-                <span class="delete-page mega-octicon octicon-trashcan" {{action 'deletePage' page}}></span>
+                <span class="delete-page mega-octicon octicon-trashcan" {{action 'deletePage' page bubbles=false}}></span>
```
Trinovantes: what does this change do?

ConnorCimowsky: Prevents the click from bubbling up to the page and showing the modal instead of just deleting it

Kapin:
Other than adding generator validation, LGTM. Please open an issue for that.

PR #152, opened 2015-06-30
Code cleanup

### LINKED ISSUE #137: Use `Ember.$` instead of `$`

ConnorCimowsky:
* Remove unnecessary gitkeep files
* Remove outdated fixture data
* Convert all indentation to 4 spaces
* Use Ember.$ instead of $

:sparkles: Resolves #137 :sparkles:


src-frontend/app/components/page-list.js
```diff
@@ -13,8 +12,8 @@ export default Ember.Component.extend({
             update: function(event, ui) {
                 var indices = {};
 
-                $(this).find('a.open-page').each(function(index) {
-                    indices[$(this).data('id')] = index;
+                Ember.$(this).find('a.open-page').each(function(index) {
```
Trinovantes: Ember has its own internal jQuery object?

Trinovantes:
+1

src-frontend/app/components/page-list.js
ConnorCimowsky: Yup, and for performance reasons you're supposed to use it instead of the global one.

ConnorCimowsky: Hm it looks like it's just an alias. But it eliminates the need for an import, so still worth using IMO

Trinovantes: sounds good

Ommy:
LGTM

PR #154, opened 2015-06-30
Fix staticfile errors

Trinovantes:
When the Rest framework tries to return an error page, it tries to load static files via the staticfiles module. Since we removed it back when we decoupled the frontend from the backend, the Rest framework can't properly serve up its error pages (e.g. 401 invalid token).

Fixes:
screenshot from 2015-06-30 10 48 51


Kapin:
LGTM

ConnorCimowsky:
:thumbsup:

PR #155, opened 2015-06-30
Add /logout endpoint and tests

Trinovantes:


src-backend/registration/tests/test_views.py
```diff
@@ -0,0 +1,23 @@
+from django.test import TestCase, Client
+from django.contrib.auth.models import User
+from rest_framework.authtoken.models import Token
+
+
+class ViewTest(TestCase):
```
Kapin: can this have a better name

Ommy:
+1 after rename of test

src-backend/registration/views.py
```diff
@@ -65,3 +92,22 @@ def login(request):
...
+# ------------------------------------------------------------------------------
+
+@api_view(['POST'])
+@authentication_classes((TokenAuthentication,))
+@permission_classes((IsAuthenticated,))
+def logout(request):
+    success = False
+
+    if request.user:
+        Token.objects.get(user=request.user).delete()
```
ConnorCimowsky:
```python
token = Token.objects.get(user=request.user)
if token:
    token.delete()
```

ConnorCimowsky:
:sparkles: LGTM after comments :sparkles:

PR #156, opened 2015-07-01
Increase generator validation strictness and associated tests

### LINKED ISSUE #150: Validate against blank/null fields in the generator

Kapin:
Closes #150

### DIFF
--- a/src-backend/api/generator.py
+++ b/src-backend/api/generator.py
@@ -8,6 +8,12 @@ class ProcedureGenerator:
         self.name = 'Procedure'
         self.procedure = procedure
 
+        if not self.procedure.title:
+            raise ValueError('Missing procedure title')
+
+        if not self.procedure.author:
+            raise ValueError('Missing procedure author')
+
     def __get_properties(self):
         props = {
             'title': self.procedure.title,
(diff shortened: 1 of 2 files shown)


src-backend/api/generator.py
```diff
@@ -8,6 +8,12 @@ def __init__(self, procedure):
         self.name = 'Procedure'
         self.procedure = procedure
 
+        if not self.procedure.title:
```
Ommy: is this testing if self.procedure.title is none? If so, why not use `if self.procedure.title is None:`

src-backend/api/generator.py
```diff
@@ -43,6 +52,24 @@ def __init__(self, element):
         self.name = 'Element'
         self.element = element
 
+        if not self.element.element_type:
+            self.__raise_error('Element with id %d has no type')
```
Ommy: All these `%d`s don't have any variable reference.

Kapin: it does. It smacks in the element id inside __raise_error

Trinovantes:
+1

Ommy:
LGTM

PR #157, opened 2015-07-01
Grant new users proper permissions

Trinovantes:
- Registration app adds the users to a default user group while the Api
  app grants that group permissions to its own models. This allows us to
  decouple the two apps.


src-backend/api/apps.py
```diff
@@ -0,0 +1,20 @@
...
+from django.contrib.contenttypes.models import ContentType
+from models import Procedure, Page, Element
+
+
+class ApiConfig(AppConfig):
+    name = 'api'
+
+    def ready(self):
+        # Grant model permissions to the appropriate user groups
+        normal_users_group, created = Group.objects.get_or_create(name='normal_users')
```
Kapin: this line doesn't make sense. It's one function call assigned to two things.

src-backend/api/apps.py
```diff
@@ -0,0 +1,20 @@
...
+        models = [Procedure, Page, Element]
+
+        for m in models:
+            content_type = ContentType.objects.get_for_model(m)
+            permissions = Permission.objects.filter(content_type=content_type)
+
+            for p in permissions:
```
Kapin: no way to just add them all en-masse?

src-backend/registration/tests/test_signup.py
```diff
@@ -0,0 +1,32 @@
...
+from django.contrib.auth.models import User
+from rest_framework.authtoken.models import Token
+import json
+
+
+class SignupTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_programmatically_created_user_has_token(self):
```
Kapin: This should be separated into a separate test class. It is a unit test vs. functional test.

src-backend/registration/tests/test_login.py
```diff
@@ -0,0 +1,40 @@
...
+import json
+
+
+class LoginTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_invalid_method_cannot_login(self):
+        response = self.client.get('/auth/login')
+        self.assertEqual(response.status_code, 405)  # Method not allowed
```
Kapin: please use the nose test stuff we have set up in the other project (or open an issue to convert it to use it)

src-backend/api/apps.py
Trinovantes: it's a list so I could just say `normal_users_group.permissions = permissions` but this would prevent other apps from adding additional permissions to this user group

src-backend/api/apps.py
Trinovantes: that function returns a tuple

src-backend/registration/tests/test_login.py
Trinovantes: good point

src-backend/registration/signals/user_post_save.py
```diff
@@ -6,10 +6,17 @@
...
+    #   rest_framework.authtoken.views.obtain_auth_token view
+    #
+    # This will create an authentication token for newly created users so the
+    # user registration endpoint can return a token back to Ember
+    # (thus avoiding the need to hit login endpoint)
     if created:
         Token.objects.create(user=instance)
+
+    # Add new user to the proper user group
+    normal_users_group, created = Group.objects.get_or_create(name='normal_users')
```
Kapin: Can we have this string pulled out somewhere?

ConnorCimowsky:
Looks good I think?

src-backend/registration/tests/test_login.py
```diff
@@ -0,0 +1,41 @@
...
+import json
+
+
+class LoginTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_invalid_method_cannot_login(self):
+        response = self.client.get('/auth/login')
+        assert_equals(response.status_code, 405)  # Method not allowed
```
Ommy: We could use http://www.django-rest-framework.org/api-guide/status-codes/ which basically will be `status.HTTP_405_METHOD_NOT_ALLOWED` instead of adding it in a comment

Ommy:
+1

src-backend/api/startup.py
```diff
@@ -0,0 +1,21 @@
+from django.conf import settings
+from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
+from models import Procedure, Page, Element
+
+
+def __grant_permissions():
```
Kapin: I'd still prefer this to be wrapped in a class, but good enough

Kapin:
LGTM

PR #159, opened 2015-07-02
Fix environment variables

ConnorCimowsky:


Kapin:
Can you make the generate env variable a generate URL and change the value in the ajax call? It's probably a little less silly and error prone.

ConnorCimowsky:
Yup agreed

ConnorCimowsky:
@Kapin err, the model's id has to be placed in the middle of the URL 😕

Kapin:
@ConnorCimowsky oh poop... yeah, don't change that then...

Trinovantes:
+1

PR #161, opened 2015-07-03
Added django extensions to project

Ommy:
It has some really useful commands such as `show_urls` which shows all routes


src-backend/sanaprotocolbuilder/settings.py
```diff
@@ -53,6 +53,7 @@
     'rest_framework.authtoken',     # Token based authentication
     'django_nose',                  # Better test framework/runner
     'corsheaders',                  # Cross-origin resource sharing
+    'django_extensions',
```
Kapin: Add a comment explaining what this does.

Kapin:
:ship: :it: :bell:

PR #164, opened 2015-07-05
Add UI for reordering element choices

ConnorCimowsky:
Coming soon: CSS and the ability to add new choices.


src-frontend/app/templates/components/choice-list.hbs
```diff
@@ -0,0 +1,8 @@
+<ul class="choices">
+    {{#each choice in choices}}
+        <li class="choice-template" choice="{{choice}}">
```
ConnorCimowsky: Note to self: this should probably be `data-choice`

src-frontend/app/components/choice-list.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+    targetObject: Em.computed.alias('parentView'),
```
Kapin: What is Em vs. Ember?

ConnorCimowsky: Ah I copied this code from an Ember doc and didn't notice. It's shorthand for `Ember`, but I agree it should be `Ember` for consistency.

ConnorCimowsky: Also in case you're wondering, this line is necessary because components send actions to the controller they reside in by default; not to their parent component if nested (which is what we want).

Kapin: Sounds good

Kapin:
LGTM based on the one comment

src-frontend/app/components/element-cell.js
```diff
@@ -0,0 +1,14 @@
...
+import SanaElement from '../models/element';
+
+export default Ember.Component.extend({
+    types: function() {
+        return SanaElement.TYPES;
+    }.property(SanaElement.TYPES),
+
+    actions: {
+        updateChoiceOrder: function(choices) {
+            this.set('model.choices', choices);
```
Trinovantes: the backend is expecting a json array right?

src-frontend/app/components/choice-list.js
```diff
@@ -0,0 +1,20 @@
...
+    didInsertElement: function() {
+        var choiceListComponent = this;
+        this.$("ul.choices").sortable({
+            handle: 'span.drag-handle',
+            revert: true,
+            scroll: true,
+            update: function(event, ui) {
+                var choices = $(this).sortable('toArray', { attribute: 'choice' });
+
+                $(this).sortable('cancel');
```
Trinovantes: is this necessary?

Trinovantes:
I will fork this branch and fix up the css later...

src-frontend/app/components/element-cell.js
ConnorCimowsky: Yup

src-frontend/app/components/choice-list.js
ConnorCimowsky: Yeah, this is what Fasih and I spent 4 hours figuring out. Otherwise jquery-sortable overrides the ordering and messes the model up.

ConnorCimowsky:
@Trinovantes Feel free to just push to this branch :)

Ommy:
+1

Trinovantes:
are you planning on changing this to the other JavaScript plugin you found? in either case, I think you should merge in this branch since I'll be fixing a lot of css later

ConnorCimowsky:
@Trinovantes I don't think so… the stuff that looked easy was actually using Angular behind the scenes. :disappointed:

ConnorCimowsky:
K merging

PR #165, opened 2015-07-06
Added in backend support to reorder pages

Ommy:
Reorder page functionality now in the backend, by calling `/api/pages/partial_bulk_update` with a JSON array of all the pages in the procedure with their new values. The method returns all the new data. 

Also added in a decorator called `initialize_once` that we can use for tests that calls the `run_once` method that initializes the group permissions for users. In the future this `run_once` will probably have more things going on.


src-backend/api/serializer.py
```diff
@@ -44,11 +44,30 @@ def to_representation(self, instance):
...
 
+class PageListSerializer(serializers.ListSerializer):
+    class Meta(object):
+        model = models.Page
+
+    def update(self, instance, validated_data):
+        page_mapping = {page.id: page for page in models.Page.objects.all()}
+        data_mapping = {item['id']: item for item in validated_data}
+
+        res = []
```
Kapin: use better name

src-backend/api/serializer.py
```diff
@@ -44,11 +44,30 @@ def to_representation(self, instance):
...
+            page = page_mapping.get(page_id, None)
+            if page is not None:
+                res.append(self.child.update(page, data))
+
+        return res
+
+
 class PageSerializer(serializers.ModelSerializer):
     elements = ElementSerializer(many=True, read_only=True)
+    id = serializers.IntegerField(read_only=False)
```
Kapin: Can we make sure that this isn't being changed somehow by the user or at least open a ticket for it?

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+from api.models import User, Page, Procedure
+from utils.decorators import add_group_permissions
+import json
+
+
+class PageTest(TestCase):
+
+    def setUp(self):
+        self.client = Client()
+        self.user = User.objects.create_user(
```
Kapin: Can you use factory boy here. If the values are irrelevant (besides associating pages and such, it'll auto-generate them.

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+            {
+                'id': self.page3.id,
+                'display_index': 2
+            }
+        ]
+
+        token = Token.objects.get(user=self.user)
+
+        response_patch = self.client.patch(
+                        path='/api/pages/partial_bulk_update/',
```
Kapin: can you make this into a constant in setup?

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+                        data=json.dumps(data),
+                        content_type='application/json',
```
Kapin: I feel like there is a constant for this somewhere

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+                        HTTP_AUTHORIZATION='Token ' + token.key)
```
Kapin: helper function?

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+                'id': self.page2.id,
+                'display_index': 0
+            },
+            {
+                'id': self.page3.id,
+                'display_index': 2
+            }
+        ]
+
+        token = Token.objects.get(user=self.user)
```
Kapin: this should also probably move into setup as well.

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+        response_patch = self.client.patch(
+                        path='/api/pages/partial_bulk_update/',
+                        data=json.dumps(data),
+                        content_type='application/json',
+                        HTTP_AUTHORIZATION='Token ' + token.key)
+
+        assert_equals(response_patch.status_code, status.HTTP_200_OK)
+        json_resp = json.loads(response_patch.content)
+
+        for page in json_resp:
```
Kapin: can you add an else assert_false(True) to ensure that nothing weird happens?

src-backend/api/tests/test_pages.py
```diff
@@ -0,0 +1,107 @@
...
+                    procedure=self.procedure
+                )
+        self.page3 = Page.objects.create(
+                    display_index=2,
+                    procedure=self.procedure
+                )
+
+    @add_group_permissions
+    def test_page_reorder_back(self):
+        data = [
```
Kapin: this also feels like it could be moved into the setup function or as a constant on the class.

src-backend/api/tests/utils/decorators.py
```diff
@@ -0,0 +1,9 @@
+import api.startup
+
+# Use this decorator to add permissions to a new user
+# created in your setUp method for tests.
+def add_group_permissions(self):
+    def _run_once():
```
Kapin: We have been using __ for private functions. Not sure if that'll cause issues here.

src-backend/api/views.py
```diff
@@ -1,8 +1,10 @@
 from django.http import HttpResponse, Http404
-from rest_framework import viewsets
-from rest_framework.decorators import detail_route
+from rest_framework import viewsets, status
+from rest_framework.response import Response
+from rest_framework.decorators import detail_route, list_route
```
Kapin: remove detail_route, it is not used.

src-backend/api/views.py
```diff
@@ -37,6 +39,33 @@ def get_queryset(self):
         user = self.request.user
         return models.Page.objects.filter(procedure__owner_id__exact=user.id)
 
+    def get_serializer(self, *args, **kwargs):
```
Kapin: Not sure what this is doing.

src-backend/api/views.py
```diff
@@ -37,6 +39,33 @@ def get_queryset(self):
...
+        if "data" in kwargs:
+            data = kwargs["data"]
+
+            if isinstance(data, list):
+                kwargs["many"] = True
+
+        return super(PageViewSet, self).get_serializer(*args, **kwargs)
+
+    @list_route(methods=["PATCH"])
```
Kapin: single quotes please

src-backend/api/views.py
```diff
@@ -37,6 +39,33 @@ def get_queryset(self):
...
+    def partial_bulk_update(self, request, *args, **kwargs):
+        partial = kwargs.pop('partial', False)
+
+        queryset = self.filter_queryset(self.get_queryset)
+
+        serializer = self.get_serializer(
+            instance = queryset,
+            data = json.loads(request.body),
```
Kapin: we should make sure the request actually has a body

src-backend/api/serializer.py
Ommy: Fine, `result`

src-frontend/bower.json
```diff
@@ -1,19 +1,19 @@
...
-        "ember-qunit-notifications": "0.0.7",
-        "ember-resolver": "~0.1.15",
-        "jquery": "^1.11.1",
-        "jquery-ui": "~1.11.4",
-        "loader.js": "ember-cli/loader.js#3.2.0",
-        "qunit": "~1.17.1",
-        "bootstrap": "~3.3.4",
-        "ember-simple-auth": "0.8.0"
-    }
+  "name": "src-frontend",
```
Kapin: Can we find a way to make this stop doing this :stuck_out_tongue:

src-backend/api/tests/test_pages.py
Ommy: Done

src-backend/api/tests/test_pages.py
Ommy: put it in `setUp`

src-backend/api/tests/test_pages.py
Ommy: Not necessary, can't find one either

src-backend/api/tests/test_pages.py
Ommy: It'll be a 2 line file but ok

`helpers.add_token_to_header(token)` is now the call. let me know if you want a better name

src-backend/api/tests/test_pages.py
Ommy: Moved

src-backend/api/tests/test_pages.py
Ommy: Different indices, not sure if we really need to do that. I tried using a json serializer, but it only worked on objects returned by `Page.objects.find(...)` so it had to be data already stored in the database. Modifying that would be a lot uglier and more work than simply defining the data as this

src-backend/api/tests/utils/decorators.py
Ommy: This is just a single `_`, reason this was done was because this seems to be the convention for decorators.

src-backend/api/views.py
Ommy: Removed when I realized that I no longer needed it since it will always be an array being return to me

src-backend/api/views.py
Ommy: Would returning a `HTTP_406_NOT_ACCEPTABLE` be what we want if the body is empty? The only time this route gets called is when we re-order, so there *should* be a body, but I can still check.

src-frontend/bower.json
Ommy: Ugh omg this happens every time. I can add the extra indentation

src-backend/api/tests/test_pages.py
Ommy: K

src-backend/api/views.py
Ommy: K

src-backend/api/tests/test_pages.py
Kapin: ok fair. I'll let the others comment here to, but that makes more sense.

src-backend/api/tests/utils/decorators.py
Kapin: works for me.

src-backend/api/tests/test_pages.py
Trinovantes: I like `assert_true(False)` better

src-backend/api/tests/utils/decorators.py
Trinovantes: There doesn't seem to be a specific convention for decorators: https://github.com/django/django/tree/8047e3666b0b50bb04e6f16c2a4fb21ddfd5713f/django/views/decorators

Some of them use `_method_name` while others use `decorator`

src-backend/api/views.py
```diff
@@ -37,6 +39,25 @@ def get_queryset(self):
         user = self.request.user
         return models.Page.objects.filter(procedure__owner_id__exact=user.id)
 
+    @list_route(methods=['PATCH'])
+    def partial_bulk_update(self, request, *args, **kwargs):
```
Trinovantes: why partial?

src-backend/api/serializer.py
```diff
@@ -44,11 +44,30 @@ def to_representation(self, instance):
         return ret
 
 
+class PageListSerializer(serializers.ListSerializer):
+    class Meta(object):
+        model = models.Page
+
+    def update(self, instance, validated_data):
+        page_mapping = {page.id: page for page in models.Page.objects.all()}
```
Trinovantes: aren't these both page mappings? if so, the variables should be renamed to better reflect where they came from e.g. `cur_mapping` and `new_mapping`

src-backend/api/tests/utils/decorators.py
```diff
@@ -0,0 +1,10 @@
+from api.startup import run_once
+
+
+# Use this decorator to add permissions to a new user
+# created in your setUp method for tests.
+def add_group_permissions(self):
```
Trinovantes: I don't like the naming for this decorator since it may one day do more than just add group permissions.

src-backend/api/views.py
```diff
@@ -37,6 +39,25 @@ def get_queryset(self):
...
+        queryset = self.filter_queryset(self.get_queryset)
+
+        if not request.body:
+            return Response(status=status.HTTP_406_NOT_ACCEPTABLE)
```
Trinovantes: I think 400 Bad Request makes more sense here

src-frontend/bower.json
```diff
@@ -15,5 +15,5 @@
         "qunit": "~1.17.1",
         "bootstrap": "~3.3.4",
         "ember-simple-auth": "0.8.0"
-    }
+      }
```
Trinovantes: the alignment in this file is off

src-backend/api/views.py
```diff
@@ -37,6 +39,25 @@ def get_queryset(self):
...
+
+        serializer = self.get_serializer(
+            instance=queryset,
+            data=json.loads(request.body),
+            many=True,
+            partial=True
+        )
+
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
```
Trinovantes: can you check if this is synchronous?

src-backend/api/tests/utils/helpers.py
```diff
@@ -0,0 +1,2 @@
+def add_token_to_header(self, token):
```
Trinovantes: may as well make this more helpful by making it take in a request object and user

src-backend/api/serializer.py
Ommy: Uh yeah I think I need to look into this one particular thing a bit more. If we have `read_only` set to `False`, then any new pages don't get saved cause now `id` is required, however, if we set `read_only` to `True`, django will assume that we don't need `id` and remove it when we do `serializer.is_valid()` which is required before proceeding with the serializer.

src-backend/api/views.py
Ommy: Partial so we don't have to pass back all the attributes of a model. Only really need the specific fields that we are updating. This will still support sending back all the attributes of a model.

src-backend/api/tests/utils/decorators.py
Ommy: Suggestions? `initialize_user_data`?

src-frontend/bower.json
Ommy: Fixed

src-backend/api/tests/utils/helpers.py
Ommy: Done

src-frontend/app/components/page-list.js
```diff
@@ -10,13 +10,19 @@ export default Ember.Component.extend({
             scroll: true,
             placeholder: 'ui-state-highlight',
             update: function(event, ui) {
-                var indices = {};
+                var sortOrder = $(this).sortable('toArray');
```
ConnorCimowsky: Forgot to mention this, but can we use `Ember.$` instead of `$`? This means you can remove the `import $` from the top

src-frontend/app/controllers/procedure.js
```diff
@@ -29,16 +29,20 @@ export default Ember.Controller.extend({
             page.save();
         },
 
-        updateSortOrder: function(indices) {
-            this.beginPropertyChanges();
+        updateSortOrder: function(pageModels) {
+            var ctx = this;
```
ConnorCimowsky: Can we call this `controller` instead of `ctx`? I think it makes it easier to understand

src-frontend/app/templates/components/page-list.hbs
```diff
@@ -1,7 +1,7 @@
 <ul class="procedure">
     <li><a class="add-page" {{action 'addPage' 0}}>+</a></li>
     {{#each page in pages}}
-        <li class="page-template">
+        <li class="page-template" id="{{page.id}}">
```
ConnorCimowsky: should probably use `data-id` here, since we're not using `id` for CSS/jQuery selectors

ConnorCimowsky: actually, I think we can get rid of this now… maybe?

ConnorCimowsky:
LGTM after comments

src-backend/api/views.py
Ommy: :+1:

src-backend/api/serializer.py
Ommy: `current_page_mapping` and `new_data_mapping` now

src-backend/api/tests/utils/decorators.py
Ommy: I'll stick with this then

src-frontend/app/components/page-list.js
Ommy: :+1:

src-frontend/app/controllers/procedure.js
Ommy: done

src-frontend/app/templates/components/page-list.hbs
Ommy: so remove?

src-backend/api/tests/test_pages.py
Ommy: K

src-backend/api/views.py
Ommy: It is. http://stackoverflow.com/questions/11899088/is-django-post-save-signal-asynchronous

Kapin:
LGTM

src-backend/api/serializer.py
```diff
@@ -44,11 +44,31 @@ def to_representation(self, instance):
...
+
+    class Meta(object):
+        model = models.Page
+
+    def update(self, instance, validated_data):
+        current_page_mapping = {page.id: page for page in instance}
+        new_data_mapping = {item['id']: item for item in validated_data}
+
+        result = []
+        for page_id, data in new_data_mapping.items():
```
Trinovantes: call this `new_page_id`

src-backend/api/tests/utils/decorators.py
```diff
@@ -0,0 +1,10 @@
+from api.startup import grant_permissions
+
+
+# Use this decorator to add permissions to a new user
+# created in your setUp method for tests.
+def initialize_permissions(self):
+    def _grant_permissions():
```
Trinovantes: can you use double `_` for our own internal consistency?

src-backend/api/views.py
Trinovantes: I'd think that bulk update is sufficient. If a field is missing, then we can assume it's unchanged.

src-frontend/app/components/page-list.js
```diff
@@ -10,13 +10,19 @@ export default Ember.Component.extend({
...
-                    indices[Ember.$(this).data('id')] = index;
+                sortOrder.forEach(function(id, index) {
+                    pageModels.push({
+                        id: id,
+                        display_index: index
+                    });
                 });
 
-                pageListComponent.sendAction('updateSortOrder', indices);
+                Ember.$(this).sortable('cancel');
```
Trinovantes: can move this to the top of this method since it's unrelated to the model update

src-frontend/app/controllers/procedure.js
```diff
@@ -29,16 +29,20 @@ export default Ember.Controller.extend({
             page.save();
         },
 
-        updateSortOrder: function(indices) {
-            this.beginPropertyChanges();
+        updateSortOrder: function(pageModels) {
+            var controller = this;
+            var onSuccess = function(data) {
```
Trinovantes: is this necessary? we should keep this in the ajax call

src-frontend/app/templates/components/page-list.hbs
Trinovantes: we're using the page id in our css rules?

src-frontend/app/controllers/procedure.js
Ommy: Eh I guess we can move it back in to the ajax call

src-frontend/app/components/page-list.js
Ommy: No. Thats not how this works. We need to get the data from the sort, but if we call cancel, it seems jquery-sortable flushes its data.

src-backend/api/tests/utils/decorators.py
Ommy: k

src-backend/api/serializer.py
Ommy: It's pretty obvious that the page_id is from the new_data_mappings but whatever, renamed

src-frontend/app/components/page-list.js
ConnorCimowsky: dat control coupling

src-backend/api/views.py
Ommy: By default, serializers must be passed values for all required fields or they will raise validation errors. You can use the partial argument in order to allow partial updates.

http://www.django-rest-framework.org/api-guide/serializers/#partial-updates

Trinovantes:
+1

PR #172, opened 2015-07-06
Allow adding new choices and editing existing choices

ConnorCimowsky:


src-frontend/app/styles/base.less
```diff
@@ -303,44 +303,120 @@ ul.element{
...
         box-sizing: border-box;
-        border-radius: 2px;
         display: block;
-        outline: none;
-        padding: 60px 60px 60px 80px;
+        line-height: @button-size;
+        text-align: center;
+
+        position: absolute;
+        left: 0; top: 0;
```
Kapin: can these be on separate lines.

src-frontend/app/styles/base.less
```diff
@@ -303,44 +303,120 @@ ul.element{
...
+        width: @button-size; height: @button-size;
```
Kapin: same as above

Kapin:
Style nit in the LESS but otherwise LGTM

Ommy:
I'd like if we could get an image of the new style in PRs, but LGTM

Trinovantes:
+1

ConnorCimowsky:
@Ommy here it is. CSS still needs to be fixed

screen shot 2015-07-06 at 4 05 34 pm

PR #173, opened 2015-07-06
Make it so each page only has a single element.

Kapin:
This change does a few things.

1) On Page creation, we now create an element for the page
2) Removes the ability to add multiple elements to a page on the front-end
3) Makes choices a non-required field (to make things behave)
4) CSS changes

### DIFF
--- a/src-frontend/app/controllers/procedure.js
+++ b/src-frontend/app/controllers/procedure.js
@@ -15,8 +15,20 @@ export default Ember.Controller.extend({
                 displayIndex: selectedIndex
             });
 
+            var procedureController = this;
+
             newPage.save().then(function() {
                 procedure.get('pages').reload();
+            }).then(function() {
+                var newElement = procedureController.store.createRecord('element', {
+                    page: newPage,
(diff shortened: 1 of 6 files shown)


src-frontend/app/controllers/procedure.js
```diff
@@ -15,8 +15,20 @@ export default Ember.Controller.extend({
                 displayIndex: selectedIndex
             });
 
+            var controller = this;
```
Trinovantes: make this `procedureController`

Ommy:
Do we still have a page has many relation with element?

ConnorCimowsky:
LGTM

src-frontend/app/controllers/procedure.js
Kapin: uggh ok

Kapin:
@Ommy yes. We are just constraining it on the client right now.

Trinovantes:
+1

Kapin:
"ehhh LGTM" -@Ommy

PR #174, opened 2015-07-07
Inject the owner automatically into created procedures

Kapin:
We no longer need to specify the owner when creating a procedure. It will be pulled out of the request and injected into the serializer.

### DIFF
--- a/src-backend/api/tests/test_procedures.py
+++ b/src-backend/api/tests/test_procedures.py
@@ -1,0 +1,36 @@
+from django.test import TestCase, Client
+from rest_framework import status
+from rest_framework.authtoken.models import Token
+from nose.tools import assert_equals, assert_true
+from utils.decorators import initialize_permissions
+from utils.helpers import add_token_to_header
+from utils import factories
+import json
+
+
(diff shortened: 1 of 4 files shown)


src-backend/api/tests/test_procedures.py
```diff
@@ -0,0 +1,36 @@
...
+        self.token = Token.objects.get(user=factories.UserFactory())
+        self.procedure_url = '/api/procedures/'
+        self.data = {
+            "title": "Example Title",
+            "author": "An Author"
+        }
+
+    @initialize_permissions
+    def test_created_procedure_has_correct_owner(self):
+        user = factories.UserFactory()
```
Ommy: Why is the user created here vs in `setUp`

Ommy:
LGTM after comments

Trinovantes:
+1

src-backend/api/tests/test_procedures.py
Kapin: user factory is a get or create. It will use a default user automatically.

ConnorCimowsky:
:+1:

PR #175, opened 2015-07-07
Added in pre-commit and pre-push

### LINKED ISSUE #127: Add commithook to indent to 4 spaces

Ommy:
Also added in a script to setup the hooks. Located in scripts/

To set up the hooks, simply go into `scripts` and run the `setup_hooks.sh` script.

Resolves #127 #84 

Edit: Realized I forgot to add the setup step into the docs, will wait until any code review changes before adding it.


Kapin:
Only comment is both of these should be pre-commit hooks, but I won't argue if the others disagree.

Ommy:
I personally don't want tests running every time i commit. Which is why we had it this way. Me an @Trinovantes discussed this.

Trinovantes:
put the hooks folder into scripts

scripts/setup_hooks.sh
```diff
@@ -0,0 +1,20 @@
...
+HOOKS_DIR="scripts/hooks"
+
+current_directory=${PWD##*/}
+
+if [ $current_directory = $SCRIPTS_DIR ]
+then
+    cd ../
+fi
+
+hooks=($(ls scripts/hooks/ | sed 's/\.sh//'))
```
ConnorCimowsky: this can be done with one command: `$(find scripts/hooks/ -iname "*.sh")`

Ommy: Yeah but that results in 
```
scripts/hooks//pre-commit.sh
scripts/hooks//pre-push.sh
```

But with the `($(ls scripts/hooks/ | sed 's/\.sh//'))` it is just the file name. The `sed` is just removing the `.sh`

ConnorCimowsky: wooooooooooooooops

Trinovantes:
+1

Kapin:
:100:

PR #181, opened 2015-07-09
Fixed re-indexing `display_index` after deletion

Ommy:
Deletion now calls `updateSortOrder` to update the display indices
New function within procedure that will generate a new page indices array.

Let me know if we should move that function somewhere else, or if we should add extra things into the `pageModel` array it returns, or a name change :dash:

### DIFF
--- a/src-frontend/app/controllers/procedure.js
+++ b/src-frontend/app/controllers/procedure.js
@@ -6,6 +6,19 @@ export default Ember.Controller.extend({
         return this.get('model.pages').sortBy('displayIndex');
     }.property('model.pages.@each.displayIndex'),
 
+    generateNewPageIndices: function() {
+        var pages = this.get('model.pages');
+        var pageModels = [];
+
+        pages.sortBy('displayIndex').forEach(function(page, index) {
+            pageModels.push({
+                id: page.id,
+                display_index: index
+            });
+        });
+        return pageModels;
+    },
+
     actions: {
         addPage: function(selectedIndex) {
             var procedure = this.get('model');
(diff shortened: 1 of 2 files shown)


src-frontend/app/controllers/procedure.js
```diff
@@ -44,17 +57,20 @@ export default Ember.Controller.extend({
         updateSortOrder: function(pageModels) {
             var controller = this;
 
+            pageModels = pageModels || controller.generateNewPageIndices()
```
Kapin: I know this is a JS idiom, but it is incredibly confusing and is missing a semicolon.

Trinovantes: I don't mind this convention since it's done on bash too. And yes, add the semicolon (how did this pass the linter?)

Is `pageModels` a global or local var? If it's global we should have a convention for them such as `m` prefix

Ommy: it's local, passed into the function on line 57

Do we even have JS linter checks on builds?

Trinovantes: If not, we should. Create issue?

Ommy: Didn't @Kapin say that everything fails when he adds JS linter to build checks? Either way, I'll create the issue

src-frontend/app/controllers/procedure.js
```diff
@@ -6,6 +6,19 @@ export default Ember.Controller.extend({
         return this.get('model.pages').sortBy('displayIndex');
     }.property('model.pages.@each.displayIndex'),
 
+    generateNewPageIndices: function() {
+        var pages = this.store.all('page');
```
ConnorCimowsky: `this.get('model.pages')` should work here

ConnorCimowsky:
looks gorm to me

Trinovantes:
+1

PR #185, opened 2015-07-14
Add fab command to start the python server.

Kapin:
Just makes it easier to run from the root folder

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -24,6 +24,11 @@ def verify():
     test()
 
 
+def runserver():
+    local('python src-backend/manage.py syncdb --noinput')
+    local('python src-backend/manage.py runserver')
+
+
 def update_host():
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -24,6 +24,11 @@ def verify():
     test()
 
 
+def runserver():
+    local('python src-backend/manage.py syncdb --noinput')
```
Ommy: Is `--noinput` just not specifying a migration file?

Kapin: uhhh, I think it just ensures that we don't have to type anything. I'm just following the rest of the fab commands.

Ommy: Yep, cool

https://docs.djangoproject.com/en/1.8/ref/django-admin/#django-admin-option---noinput

Ommy:
lgtm

PR #186, opened 2015-07-14
Fixed addition of pages at specific indexes

### LINKED ISSUE #182: Adding a new page should push the pages below it down and increment their indices
Currently it gets assigned the current index where the `+` was clicked, but doesn't update the display indices of the rest of the pages below that `+`. We should increment all the other pages and send a bulk update.

Ommy:
This is when we want to create a new page in between two pages.
Before it wasn't updating the display indices of the other pages.

Resolves #182

### DIFF
--- a/src-frontend/app/controllers/procedure.js
+++ b/src-frontend/app/controllers/procedure.js
@@ -22,6 +22,18 @@ export default Ember.Controller.extend({
     actions: {
         addPage: function(selectedIndex) {
             var procedure = this.get('model');
+            var pagesToUpdate = [];
+
+            this.get('model.pages').forEach(function(page) {
+                if (page.get('displayIndex') >= selectedIndex) {
+                    pagesToUpdate.push({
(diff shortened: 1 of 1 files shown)


Kapin:
LGTM

Trinovantes:
Is this handled on the server too?

ConnorCimowsky:
Should we be updating the records locally before sending the request? I think right now we only do this when we receive the response and push the records into the store.

Trinovantes:
Yea I'd rather let the server notify Ember that some page changed on the backend than to let Ember guess what's happening on the backend

ConnorCimowsky:
Let's talk about this in person tomorrow :+1:

ConnorCimowsky:
Re: offline discussion, I think we'll keep it as-is?

Ommy:
"+1" - @Trinovantes

ConnorCimowsky:
+1

PR #187, opened 2015-07-14
Add a button to the procedure cards to delete them.

Kapin:
This will allow us to delete existing procedures.

Closes #177

PR #188, opened 2015-07-14
Update and handle last_modified timestamps correctly.

Kapin:
Depends on #187 because of UI.

Does two things.

1. Fixes updating last_modified timestamps for models on the server so that child elements (pages, elements) cascade their updates to their parents.
2. Makes it so that procedures use their actual last_modified timestamp and are sorted in the correct order.

### DIFF
--- a/src-backend/api/tests/test_model_page.py
+++ b/src-backend/api/tests/test_model_page.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 from django.contrib.auth.models import User
 from django.db import IntegrityError
-from nose.tools import raises, assert_equals, assert_not_equals, nottest
+from nose.tools import raises, assert_equals, assert_not_equals, assert_true, nottest
 from api.models import Procedure, Page
 from utils import factories
 
@@ -74,4 +74,4 @@ class ProcedureTest(TestCase):
             page=page
         )
 
-        assert_not_equals(original_last_modified, page.last_modified)
+        assert_true(original_last_modified < page.last_modified)
--- a/src-backend/api/tests/test_model_page.py
+++ b/src-backend/api/tests/test_model_page.py
@@ -3,6 +3,7 @@ from django.contrib.auth.models import User
 from django.db import IntegrityError
 from nose.tools import raises, assert_equals, assert_not_equals, nottest
 from api.models import Procedure, Page
+from utils import factories
 
 
 class ProcedureTest(TestCase):
@@ -64,3 +65,13 @@ class ProcedureTest(TestCase):
             display_index=0,
             procedure=self.test_procedure1
         )
+
+    def test_updates_last_modified(self):
+        page = factories.PageFactory()
+        original_last_modified = page.last_modified
+
+        factories.ElementFactory(
+            page=page
+        )
+
+        assert_not_equals(original_last_modified, page.last_modified)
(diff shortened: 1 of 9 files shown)


src-backend/api/tests/test_model_procedure.py
```diff
@@ -71,3 +72,13 @@ def test_owner_none(self):
...
+
+    def test_updates_last_modified(self):
+        procedure = factories.ProcedureFactory()
+        original_last_modified = procedure.last_modified
+
+        factories.PageFactory(
+            procedure=procedure
+        )
+
+        assert_not_equals(original_last_modified, procedure.last_modified)
```
Trinovantes: should assert current `last_modified` is greater than `original_last_modified`

src-backend/api/tests/test_model_page.py
```diff
@@ -64,3 +65,13 @@ def test_display_index_uniqueness(self):
...
+
+    def test_updates_last_modified(self):
+        page = factories.PageFactory()
+        original_last_modified = page.last_modified
+
+        factories.ElementFactory(
+            page=page
+        )
+
+        assert_not_equals(original_last_modified, page.last_modified)
```
Trinovantes: here too

Trinovantes:
shouldn't modifying elements/choices change last modified too?

src-backend/api/tests/test_model_procedure.py
Kapin: will try

src-backend/api/tests/test_model_page.py
Kapin: see above

Kapin:
@Trinovantes it should cascade, but I'll verify.

src-frontend/app/controllers/procedures/index.js
```diff
@@ -1,6 +1,10 @@
 import Ember from 'ember';
 
 export default Ember.Controller.extend({
+    procedures: function() {
+        return this.get('model').sortBy('lastModified').reverse();
```
Ommy: Wish there was a cleaner way to do this.

Kapin: @ConnorCimowsky probably knows a better way

src-frontend/app/helpers/date-string.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
```
Ommy: Are we really going to have a new file for every handlebar helper function?

ConnorCimowsky: Yup, this is the convention

src-frontend/app/controllers/procedures/index.js
ConnorCimowsky: Would this work?

```javascript
procedures: function() {
  return Ember.ArrayProxy.createWithMixins(Ember.SortableMixin, {
    sortProperties: ['lastModified'],
    sortAscending: false,
    content: this.get('model')
  });
}.property('model.@each.lastModified')
```

ConnorCimowsky:
LGTM after comments

Ommy:
+1 after comments

src-frontend/app/controllers/procedures/index.js
Kapin: I'll try that

Trinovantes:
+1

PR #190, opened 2015-07-15
Transition to procedures if logged in

### LINKED ISSUE #179: Going to the website should redirect to /procedures/ if the user is logged in.
Currently it goes to a blank screen. The default view should be the procedure view.

screen shot 2015-07-08 at 3 10 42 pm

ConnorCimowsky:
Resolves #179

### DIFF
--- a/src-frontend/app/routes/index.js
+++ b/src-frontend/app/routes/index.js
@@ -1,4 +1,7 @@
 import Application from 'src-frontend/routes/application';
 
 export default Application.extend({
+    beforeModel: function() {
+        this.transitionTo('procedures');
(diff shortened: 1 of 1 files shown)


Ommy:
Does this guarantee that the user is logged in before transitioning? Like if we get to that `beforeModel`, we know the user is logged in?

ConnorCimowsky:
No, but the procedures route will redirect to the login page if they're not logged in. Is that the behaviour we want?

Ommy:
Yeah that works, +1

Kapin:
:boom:

PR #192, opened 2015-07-17
Added in flash messages

### LINKED ISSUE #123: Add flash messages to transitions for things like sign ups, form validations
https://github.com/ryanto/ember-flash-message

Ommy:
I've used https://github.com/poteto/ember-cli-flash to implement the flash messages. So far we only have it in the sign up page, but @ConnorCimowsky can add them into the login flow (I'm not 100% sure how to work the auth flow so I'll leave that to Connor).

The CSS changes are because by default the flash message takes up pretty much 100% of the screen, which doesn't look very good.

Resolves #123 

Screenshot: 

screen shot 2015-07-17 at 4 04 53 pm

TODO: Add form validations and add the errors into the flash messages (i.e. missing fields). Will do in another ticket

### DIFF
--- a/src-frontend/app/controllers/signup.js
+++ b/src-frontend/app/controllers/signup.js
@@ -11,6 +11,7 @@ export default Ember.Controller.extend({
     actions: {
         signup: function() {
             var controller = this;
+            var flashMessages = Ember.get(this, 'flashMessages');
             Ember.$.post(ENV.APP.API_HOST + '/auth/signup', {
                 email: this.get('email'),
                 username: this.get('username'),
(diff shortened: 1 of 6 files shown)


Kapin:
screenshot plz. otherwise lgtm

Kapin:
:trumpet: :trumpet:

src-frontend/app/templates/application.hbs
```diff
@@ -1,2 +1,5 @@
+{{#each flashMessages.queue as |flash|}}
+  {{flash-message flash=flash}}
```
ConnorCimowsky: 4-space indent :trollface:

ConnorCimowsky:
:+1:

src-frontend/app/controllers/signup.js
```diff
@@ -19,11 +20,10 @@ export default Ember.Controller.extend({
                 accept_tos: this.get('accept_tos')
             }, function(response) {
                 if (response.success === true && response.token) {
+                    flashMessages.success('Account was created successfully!');
```
Trinovantes: This should use the error message returned from the server

PR #194, opened 2015-07-19
Get Ember tests to run; add them to Fabric/Travis

### LINKED ISSUE #166: Travis should run `ember test` on all builds
This will run the front-end tests and lint the js files. We already do this on the backend so it is a good practice.

ConnorCimowsky:
:information_source:  Resolves #166

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,11 @@ addons:
 branches:
   only:
     - master
+before_install:
+  - "npm install -g npm@^2"
 install:
   - pip install -r requirements.txt
+  - bash -x ./scripts/install-frontend.sh
 before_script:
   - createuser --superuser test_db_user -U postgres
   - psql -c "ALTER USER test_db_user WITH PASSWORD 'test_db_password';" -U postgres
--- a/scripts/install-frontend.sh
+++ b/scripts/install-frontend.sh
@@ -1,0 +1,4 @@
+cd src-frontend
+npm install -g bower
+npm install
+bower install
--- a/fabfile.py
+++ b/fabfile.py
@@ -14,6 +14,9 @@ def test():
     local('python src-backend/manage.py test api --noinput')
     local('python src-backend/manage.py test registration --noinput')
 
+    with lcd('src-frontend'):
+        local('ember test')
+
 
 def lint():
     local('flake8 src-backend')
--- a/src-frontend/.watchmanconfig
+++ b/src-frontend/.watchmanconfig
@@ -1,0 +1,3 @@
+{
+  "ignore_dirs": ["tmp"]
+}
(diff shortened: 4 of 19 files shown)


Kapin:
LGTM

ConnorCimowsky:
This actually doesn't work as of now, I submitted the PR so I could see what happens with Travis. I'll let y'all know when it's ready for review, since it'll likely be more complicated than it currently is. :hankey:

Kapin:
I just saw the travis changes now. If you need help poking around at it we can sit down and sort it out.

ConnorCimowsky:
Cool, thanks! I think I might almost have it working, but if it doesn't we can pair on Monday

.travis.yml
```diff
@@ -10,8 +10,11 @@ addons:
 branches:
   only:
     - master
+before_install:
+  - "npm install -g npm@^2"
```
Kapin: The node versioning syntax still makes absolutely no sense :confounded:

scripts/install-frontend.sh
```diff
@@ -0,0 +1,4 @@
+cd src-frontend
+npm install -g bower
+npm install
+bower install
```
Kapin: nit, but can this go back to the previous directory when done? I feel like it won't matter all that much, but it is good to cleanup.

Kapin:
Only question is does ember test run jshint as well? I think it's good practice for us to run it on every build as well.

Kapin:
Also can you add this to the pre-commit hooks and the fab test command?

scripts/install-frontend.sh
ConnorCimowsky: That's actually why I made this a separate script for `bash -x`. It won't affect the environment of the caller.

ConnorCimowsky:
Yes this runs JSHint :thumbs:

ConnorCimowsky:
Already in `fab test` but I'll add it to the hook

scripts/install-frontend.sh
Kapin: perfect

Kapin:
LGTM once the hook is in.

scripts/install-frontend.sh
Ommy: If I call this while within `scripts/` then won't `cd src-frontend` fail?

Ommy:
LGTM after comments + hooks.

Also, would you mind squashing some of those commits?

Trinovantes:
+1

scripts/install-frontend.sh
ConnorCimowsky: How do you mean? Is there a non-gross way of doing this?

ConnorCimowsky:
The Ember tests run in `fab test` which runs in `fab verify`, which runs in the pre-push hook, so we're good. Will squash and merge.

PR #193, opened 2015-07-19
Get Ember tests to pass

ConnorCimowsky:
Plus, add some unit tests for `model:element`

PR #195, opened 2015-07-20
Remove sudo's and cleanup start scripts

Kapin:
Probably should wait until #193 is merged in, but this cleans up some things and makes setup scripts not use sudo.

### DIFF
--- a/scripts/dev_setup_linux.bash
+++ b/scripts/dev_setup_linux.bash
@@ -6,7 +6,7 @@ sudo apt-get update
 # Install python, pip
 sudo apt-get install -y python
 sudo apt-get install -y python-pip
-sudo pip install -r ../requirements.txt
+pip install -r ../requirements.txt
 
 # Install Postgres
 sudo apt-get install -y postgresql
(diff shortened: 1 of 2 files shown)


src-frontend/package.json
```diff
@@ -37,6 +37,8 @@
         "ember-disable-proxy-controllers": "^0.7.0",
         "ember-django-adapter": "^0.5.6",
         "ember-export-application-global": "^1.0.2",
-        "ember-git-version": "0.0.1"
+        "ember-git-version": "0.0.1",
+        "grunt-cli": "^0.1.13",
+        "less": "^2.5.1"
```
ConnorCimowsky: don't need `less` here, since we already require `ember-cli-less`

ConnorCimowsky:
LGTM after comment, let's wait on #194 before merging

src-frontend/package.json
Kapin: k

Trinovantes:
+1

Ommy:
lgtm

ConnorCimowsky:
bump now that #194 is merged :stars:

PR #196, opened 2015-07-20
Add element unit tests and refactor page and procedure unit tests

### LINKED ISSUE #101: Ensure models have tests
Need to catch up on this.

Kapin:
Gives some validation for the element models and cleans up the page/procedure model tests a lot.
Additionally, covers some missing cases from procedures and pages as well.

Also fixes a bug that was found in creating the element tests. The server was not validating the type of the element properly and that is now fixed.

Closes #101

### DIFF
--- a/src-backend/api/tests/test_model_element.py
+++ b/src-backend/api/tests/test_model_element.py
@@ -1,0 +1,90 @@
+from django.test import TestCase
+from django.db import IntegrityError
+from nose.tools import raises, assert_equals, assert_not_equals
+from api.models import Element
+from utils import factories
+
+
+class ElementTest(TestCase):
+    def test_elements_required(self):
+        page = factories.PageFactory()
+        element = Element.objects.create(
+            display_index=0,
+            page=page
+        )
+
+        assert_equals(element.display_index, 0)
+        assert_equals(element.page, page)
+
+    def test_elements_all_properties(self):
+        page = factories.PageFactory()
+        Element.objects.create(
+            display_index=0,
+            eid='eid',
+            element_type='SELECT',
+            choices='[one]',
+            numeric='DIALPAD',
+            concept='TEST',
+            question='test question',
+            answer='',
+            page=page
+        )
+
+        element = Element.objects.get(concept='TEST')
+
+        assert_equals(element.display_index, 0)
+        assert_equals(element.eid, 'eid')
+        assert_equals(element.element_type, 'SELECT')
+        assert_equals(element.choices, '[one]')
+        assert_equals(element.numeric, 'DIALPAD')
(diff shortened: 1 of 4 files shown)


Ommy:
Nice, factories. Looks cleaner. +1

src-backend/api/tests/test_model_element.py
```diff
@@ -0,0 +1,109 @@
...
+            answer='',
+            page=page
+        )
+
+        element = Element.objects.get(concept='TEST')
+
+        assert_equals(element.display_index, 0)
+        assert_equals(element.eid, 'eid')
+        assert_equals(element.element_type, 'SELECT')
+        assert_equals(element.choices, '[one]')
```
Trinovantes: maybe use a json decoder instead of hardcoding the json string?

src-backend/api/tests/test_model_element.py
```diff
@@ -0,0 +1,109 @@
...
+        assert_equals(element.numeric, 'DIALPAD')
+        assert_equals(element.concept, 'TEST')
```
Trinovantes: make this `TESTCONCEPT` to make it easier to search for unique strings

src-backend/api/tests/test_model_element.py
Kapin: It's irrelevant, it just stores it as a string.

src-backend/api/tests/test_model_element.py
Kapin: k

src-backend/api/tests/test_model_element.py
Trinovantes: it's just a test convention

at the very least, we don't have to worry about some test failing down the line because the setup had a typo

Kapin: This is a test to explicitly ensure the choices is stored as a string properly. I don't want to combine concerns here. We have other tests that cover the json portions.

Trinovantes:
These two commits seem disjoint and should be split into 2 PRs

Kapin:
@Trinovantes the second commit fixes the fact that the indenting is messed up on element tests. Additionally, it found a bug in elements where we don't actually validate choices properly, and hence fixes that.

Trinovantes:
Why do you need 1 commit to fix another commit in the same PR?

Kapin:
I can squash them. Can we focus on the code versus arguing over that I used two commits?

ConnorCimowsky:
dis-gon-b-gud-oh-man

Ommy:
lgtm

Trinovantes:
+1

Trinovantes:
Can we get this merged in yet?

PR #201, opened 2015-07-27
Clean up some of the create page UX

### LINKED ISSUE #178: If a user cancels page creation, it should delete the element
We should not be persisting a page if we press cancel.

Kapin:
This change moves the logic for adding an element to a page automatically to the edit functionality of the page creation versus on creation. It also changes "cancel" to "close" to make it more explicit what it is doing.

Sorta resolves #178 in a different way.


Trinovantes:
+1

Ommy:
LGTM

ConnorCimowsky:
LGTM

PR #202, opened 2015-07-28
Added checks for failed creations of page/procedure

### LINKED ISSUE #145: Create element/page/procedure should first check if server returned 200 before updating UI
If the server does not return a 200 when trying to create a new element/page/procedure, the UI still updates with a new component with id of null. It should instead show a loading spinner and wait for a server response before updating the UI.

Ommy:
When creating a new page or procedure, fails on the save will destroy the object instead. Flash Messages will also be displayed. Resolves #145


src-frontend/app/controllers/procedure.js
```diff
@@ -19,9 +19,20 @@ export default Ember.Controller.extend({
         return pageModels;
     },
 
+    generateErrorString: function(errors, type) {
+        var message = 'Something went wrong creating the ' + type;
+        errors.forEach(function(error){
+            for (var attr in error.details) {
+                message = attr.toString() + ": " + error.details[attr][0];
```
Kapin: shouldn't this just be attr[0] not error.details[attr][0]?

Kapin: also single quotes pls

src-frontend/app/controllers/procedures/index.js
```diff
@@ -20,6 +21,15 @@ export default Ember.Controller.extend({
 
             procedure.save().then(function() {
                 procedureController.transitionToRoute('procedure', procedure);
+            }).catch(function(response) {
+                var errorMessage = 'Something went wrong trying to create the procedure';
+                response.errors.forEach(function(error){
+                    for (var attr in error.details) {
+                        errorMessage = attr.toString() + ": " + error.details[attr][0];
```
Kapin: see above

src-frontend/app/controllers/procedures/index.js
```diff
@@ -20,6 +21,15 @@ export default Ember.Controller.extend({
 
             procedure.save().then(function() {
                 procedureController.transitionToRoute('procedure', procedure);
+            }).catch(function(response) {
+                var errorMessage = 'Something went wrong trying to create the procedure';
```
Kapin: this should call generateErrorString

src-frontend/app/controllers/procedure.js
Ommy: No, cause attr will be the attribute such as `display_index`, so we access the value of that message like if it was `display_index: "Can't be blank!"`

Kapin: :skull: :gun:

src-frontend/app/controllers/procedures/index.js
Ommy: I can't call that function from within this controller. @ConnorCimowsky is there some Ember convention for putting JS helper functions (not for Handlebars)

src-frontend/app/controllers/procedure.js
Trinovantes: It should be double quotes for UI strings. We only use single quotes for key-like strings

src-frontend/app/controllers/procedure.js
```diff
@@ -54,6 +65,10 @@ export default Ember.Controller.extend({
                 newElement.save().then(function() {
                     procedureController.send('editPage', newPage);
                 });
+            }).catch(function(response) {
+                var errorMessage = procedureController.generateErrorString(response.errors, 'page');
+                flashMessages.danger(errorMessage);
+                newPage.destroy();
```
Trinovantes: does this route us back to the state before the page is created?

Trinovantes:
Stopped development with Ember

PR #206, opened 2015-07-28
Add undefined check before rendering a procedure date string

Kapin:
Weird race condition is causing a console error due to the last modified
date not being set by the server immediately. This fixes that.


Trinovantes:
When it eventually gets set, does the UI get updated?

Kapin:
That will never happen as we transition to the next page. I have no found a way to test it.

Trinovantes:
Why does this old UI get updated if we're supposed to transition?

Kapin:
@Trinovantes it's a race condition with ember. The model is created before the transition occurs. This is a much simpler fix than forcibly preventing a refresh from ember.

Trinovantes:
Can you ask the developers on IRC to see if this the only solution or if there's another paradigm that we're not aware of?

Trinovantes:
Stopped development with Ember

PR #221, opened 2015-10-27
Setup Backbone testing with mocha, chai, sinon

### LINKED ISSUE #213: Setup backbone testing framework

Trinovantes:
The tests are headless so they need to load JS libs via require (loaded
from npm repository). As a result, we are no longer using libraries
loaded from bower. Those loaded from bower cannot be (easily) tested.

I also updated our package.json to remove some warnings.

Closes #213

### DIFF
--- a/src-backbone/tests/smoke_test.js
+++ b/src-backbone/tests/smoke_test.js
@@ -1,0 +1,37 @@
+var CreateSuite = require('utils/create_suite');
+
+CreateSuite('Smoke Test', function() {
+    describe('Array#indexOf', function () {
+        it('should return -1 when the value is not present', function () {
+            assert.equal(-1, [1, 2, 3].indexOf(4));
+        });
+    });
+
+    describe('SanaApp', function() {
+        beforeEach(function() {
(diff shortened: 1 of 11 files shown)


ConnorCimowsky:
:thumbs:

src-backbone/tests/smoke_test.js
```diff
@@ -0,0 +1,37 @@
+var CreateSuite = require('utils/create_suite');
+
+CreateSuite('Smoke Test', function() {
+    describe('Array#indexOf', function () {
+        it('should return -1 when the value is not present', function () {
+            assert.equal(-1, [1, 2, 3].indexOf(4));
```
Ommy: Why?

Ommy:
LGTM once comment is addressed

src-backbone/tests/smoke_test.js
Trinovantes: To make sure our tests can run at all without depending on our main app. ¯\_(ツ)_/¯

PR #222, opened 2015-10-29
Wrote tests for authentication module

### LINKED ISSUE #214: Write tests for Backbone auth system
Depends on #213

Trinovantes:
Also refactored exisiting code with dependency injection so that they're
easier to test.

Closes #214

### DIFF
--- a/src-backbone/app/js/routers/auth/auth_controller.js
+++ b/src-backbone/app/js/routers/auth/auth_controller.js
@@ -1,42 +1,44 @@
-var Helpers = require('utils/helpers');
-
 module.exports = Marionette.Controller.extend({
 
+    constructor : function (options) {
+        this.app = options.app;
+        this.Helpers = (options && options.Helpers) || require('utils/helpers');
+        this.AuthLayout = (options && options.AuthLayout) || require('views/auth/auth_layout');
+        this.SignupView = (options && options.SignupView) || require('views/auth/signup_view');
+        this.LoginView = (options && options.LoginView) || require('views/auth/login_view');
+
+        Marionette.Controller.prototype.constructor.call(this, options);
+    },
+
     signup: function () {
-        if (App.session.isValid()) {
-            Helpers.goto_default_logged_in();
+        if (this.app.session.isValid()) {
+            this.Helpers.goto_default_logged_in();
             return;
         }
 
-        console.info('View: Sign Up');
+        log.info('View: Sign Up');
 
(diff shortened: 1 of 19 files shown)


src-backbone/app/js/routers/auth/auth_controller.js
```diff
@@ -1,42 +1,44 @@
-var Helpers = require('utils/helpers');
-
 module.exports = Marionette.Controller.extend({
 
+    constructor : function (options) {
+        this.app = options && options.app;
```
Trinovantes: technically for here and all other instances of `this.app = options && options.app;` I can skip the null check on `options` to enforce the presence of `app` passed into the constructor.

ConnorCimowsky:
LGTM

src-backbone/app/js/routers/auth/auth_controller.js
Ommy: Btw in Coffeescript this would just be `options?.app`

src-backbone/tests/auth/auth_controller_test.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+
+        helpers_mock = sinon.mock(auth_controller.Helpers);
+    });
+
+    afterEach(function() {
+        helpers_mock.verify();
+    });
+
+    describe("signup()", function() {
```
Ommy: Nit but describes for functions should have a hash at the front, i.e. `"#signup()`

src-backbone/tests/auth/session_model_test.js
```diff
@@ -0,0 +1,229 @@
...
+
+            session._auth_handler(response, server_error_handler);
+
+            assert.strictEqual(session.get(SessionModel.AUTH_TOKEN_KEY), TOKEN);
+            assert(save_spy.calledOnce);
+            assert(server_error_handler.notCalled);
+        });
+
+        it("should call server_error_handler if request was unsuccessful on the backend", function() {
+            var ERROR_MESSAGE = 'error';
```
Ommy: Any reason why this is initialized like a constant?

Trinovantes: it should be a local const but I don't think there's a convention for that?

src-backbone/app/js/routers/auth/auth_controller.js
Trinovantes: grrr >:(

Issue #212, opened 2015-10-23
Minor chores before everyone starts working again


Trinovantes:
- [x] Delete `mockups` folder
- [x] Update wiki and delete `docs` folder
- [x] Move `requirements.txt` into `src-backend` folder
- [x] Rename `registration` module to `authentication` module in Django 
- [ ] Rename `src-backbone` to `src-frontend` after everything is done

PR #229, opened 2015-10-29
Clean up authentication

### LINKED ISSUE #211: Create Heartbeat endpoint to check if token is valid
We need a `/heartbeat` endpoint to check if a user token is valid. It should be a barebones Django view that returns an empty body. It should return 200 for valid token or 401 for invalid token.

Trinovantes:
* Renamed `registration` module in Django to `authentication` to keep it similar to the frontend `auth` module.
* Removed heartbeat check.

`Session.isValid()` is being called a lot by the UI so making it
synchronous would be bad UX. But if we make it an async call, then it
would more or less be the same as using an invalid token to fetch a
model/collection down the line since both would eventually clear the
client Session model and redirect the user to `/`

Closes #211, Part of #212


Kapin:
:beer:

Ommy:
+1

PR #231, opened 2015-10-29
Delete docs folder

Trinovantes:
Moving documentation to wiki so we don't need a pull request everytime
we want to make a minor change.

Part of #212


Kapin:
:1234:

Ommy:
:+1:

PR #235, opened 2015-11-04
Implemented procedures list

Trinovantes:
Can't wait till we hit those JavaScript race conditions

I would suggest reading the commits separately. Tests will come after the port is finished (or someone else can write them).


Trinovantes:
screenshot from 2015-11-04 17 31 48
screenshot from 2015-11-04 17 31 28

src-backbone/app/js/views/procedures/account_navbar_view.js
```diff
@@ -0,0 +1,21 @@
...
+        Marionette.ItemView.prototype.constructor.call(this, options);
+    },
+
+    templateHelpers: function() {
+        var self = this;
+
+        return {
+            username: function() {
+                // TODO fetch from server
+                return 'Trinovantes';
```
Ommy: So I guess this will be a separate ticket?

Ommy:
Overall looks fine. I just don't like how the colour for the background saying you don't have any procedures is the same as the list of procedures. I think if a user has no procedures, something else should be there, not a list element. Maybe a static image saying they don't have procedures, or simply a http://getbootstrap.com/components/#alerts.

Trinovantes:
Maybe yellow? Then again, it's just a note and not a warning or error. I could try making the procedures lighter colour

src-backbone/app/js/views/procedures/account_navbar_view.js
Trinovantes: Yea we need an `/api/account` endpoint for getting info and changing email/passwords

Ommy:
I guess maybe that bootstrap link is not the best suggestion, but there should be some difference between an empty procedure list and a populated one.

Trinovantes:
make procedures blue?

Ommy:
Yeah maybe that'll work. Try it out. Just some difference between the two will be good.

Trinovantes:
The blues look too harsh under the grey text. I decided to just keep it as a white bordered card with rounded corners.

screenshot from 2015-11-06 17 51 46

Ommy:
Is an empty list of procedures still the gray box?

Trinovantes:
Yea

Trinovantes:
screenshot from 2015-11-09 17 19 27
screenshot from 2015-11-09 17 19 12

Ommy:
New design looks slick, +1

src-backbone/app/js/models/procedures_collection.js
```diff
@@ -0,0 +1,57 @@
...
+        return (this._sort_asc ? 1 : -1) * this._asc_comparator(a.get(this._comparator_key), b.get(this._comparator_key));
+    },
+
+    set_comparator_key: function(key) {
+        log.info('Procedure Comparator Key: ' + key);
+        this._comparator_key = key;
+    },
+
+    set_order_asc: function(sort_asc) {
+        log.info('Procedure Sort Asc: ' + sort_asc);
```
ConnorCimowsky: Do we want to keep these log statements?

src-backbone/app/js/models/procedures_model.js
```diff
@@ -0,0 +1,19 @@
+module.exports = Backbone.Model.extend({
+
+    defaults: {
+        'title': 'Untitled', // TODO allow nullable titles
+        'author': 'Stephen', // TODO either fetch user's username or let API allow nullable authors
+        'pages': [],
+    },
+
+    parse: function(response, options) {
+        var parse_date_in_response = function (key) {
```
ConnorCimowsky: epic nit but I don't think the space is intentional in `function (key)`

ConnorCimowsky:
LGTM :1:

src-backbone/app/js/models/procedures_collection.js
Trinovantes: These output to the browser's console to help us debug. I don't think we actually need to keep these in production (maybe we can replace it with analytics events?)

PR #241, opened 2015-11-20
Cleanup frontend

Trinovantes:
A lot of small changes that I found while going through the source code.

- Removed coffeescript since we're already mostly JS
- Use ES6 to clean up some code (e.g. default function param)
- Setup browserify in gulp task "the right way". This includes compiling
  Handlebars templates with `hbsfy`. This also allows us to extend our test
  coverage to views since the `templates` attribute can now be handled with
  dependency injection.
- Localized gulp packages being loaded to avoid polluting the global
  namespace in `gulpfile.js`
- Updated logger initialization so that it's cleaner
- Updated `API_BASE` in Config file to include `/api`
- Moved `DEBUG` flag to Config file
- Added empty files for builder


src-backbone/app/js/templates/procedures/builder_layout.hbs
```diff
@@ -0,0 +1 @@
+hello world!
```
Ommy: Required?

Ommy:
RIP in F CoffeeScript :cry: 

LGTM after comment

src-backbone/app/js/views/procedures/procedures_layout.js
```diff
@@ -78,7 +78,9 @@ module.exports = Marionette.LayoutView.extend({
         event.preventDefault();
 
         var self = this;
-        var new_procedure = new this.Procedure({}, {
+        var new_procedure = new this.Procedure({
+            author: 'Stephen' // TODO either fetch user's username or let API allow nullable authors
```
ConnorCimowsky: just making sure this is intentional

ConnorCimowsky:
rest in pepperoni, coffeescript

ConnorCimowsky:
LGTM :sparkles:

src-backbone/app/js/views/procedures/procedures_layout.js
Trinovantes: Yea this is actually the correct way to initialize a model. I was actually surprised the previous way worked too

PR #243, opened 2015-11-26
Fixed FOUC when procedures are not sorted

Trinovantes:
If Marionette detects the the initial collection is not sorted properly,
it will re-render the CollectionView on page load.


Ommy:
+1

PR #260, opened 2015-12-08
Updated procedure API to include only_return_id

### LINKED ISSUE #256: /api/procedures return page id instead of page contents
e.g. When I make a request to `localhost:8000/api/procedures/100`, I get:
```
{
  "id": 100,
  "title": "Untitled Procedure 2",
  "author": "Stephen",
  "version": null,
  "uuid": null,
  "owner": 3,
  "pages": [
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9,
    10,
    11,
    12,
    13,
    14,
    15
  ],
  "last_modified": "2015-12-06T23:47:15.317210Z",
  "created": "2015-12-06T21:50:40.178327Z"
}
```

If the page id is specified, it should transmit the page contents instead of just the primary keys. 

Note that if the page is not specified (`/api/procedures`), then we can discard the `pages` attribute since it's not needed (alternatively, make it optional via a flag e.g. `?return_page_contents=true`)

Ommy:
only_return_id flag for procedures, when set true, will result in a
response with only page ids for each procedure. If however it is set
to false, then instead of page ids, the full page object will be returned.

Resolves #256

### DIFF
--- a/src-django/api/tests/test_procedures.py
+++ b/src-django/api/tests/test_procedures.py
@@ -9,28 +9,48 @@ import json
 
 
 class ProcedureTest(TestCase):
+
     def setUp(self):
         self.client = Client()
         self.token = Token.objects.get(user=factories.UserFactory())
-        self.procedure_url = '/api/procedures/'
+        self.procedure_url = '/api/procedures'
         self.data = {
-            "title": "Example Title",
-            "author": "An Author"
+            'title': 'Example Title',
+            'author': 'An Author'
         }
+        self.user = factories.UserFactory()
+        initialize_permissions
 
-    @initialize_permissions
     def test_created_procedure_has_correct_owner(self):
-        user = factories.UserFactory()
-
         response = self.client.post(
             path=self.procedure_url,
             data=json.dumps(self.data),
             content_type='application/json',
-            HTTP_AUTHORIZATION=add_token_to_header(user, self.token)
+            HTTP_AUTHORIZATION=add_token_to_header(self.user, self.token)
+        )
+
+        assert_equals(response.status_code, status.HTTP_201_CREATED)
+        body = json.loads(response.content)
+
+        assert_true('owner' in body)
+        assert_equals(body['owner'], self.user.id)
+
+    def test_procedure_comes_with_pages(self):
+        procedure = factories.ProcedureFactory(owner=self.user)
+        factories.PageFactory(procedure=procedure)
+        flags = {
+            'only_return_id': 'false'
+        }
+
+        response = self.client.get(
+            path=self.procedure_url,
+            data=flags,
+            content_type='application/json',
+            HTTP_AUTHORIZATION=add_token_to_header(self.user, self.token)
         )
 
         assert_equals(response.status_code, status.HTTP_200_OK)
         body = json.loads(response.content)
-
-        assert_true('owner' in body)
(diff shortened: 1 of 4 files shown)


src-django/api/tests/test_procedures.py
```diff
@@ -9,28 +9,48 @@
...
 
         assert_true('owner' in body)
-        assert_equals(body['owner'], user.id)
+        assert_equals(body['owner'], self.user.id)
+
+    def test_procedure_comes_with_pages(self):
+        procedure = factories.ProcedureFactory(owner=self.user)
+        factories.PageFactory(procedure=procedure)
+        flags = {
+            "only_return_id": "false"
```
Trinovantes: Can you use single quotes here?

src-django/api/tests/test_procedures.py
```diff
@@ -9,28 +9,48 @@
...
-    @initialize_permissions
     def test_created_procedure_has_correct_owner(self):
-        user = factories.UserFactory()
-
         response = self.client.post(
             path=self.procedure_url,
             data=json.dumps(self.data),
             content_type='application/json',
-            HTTP_AUTHORIZATION=add_token_to_header(user, self.token)
+            HTTP_AUTHORIZATION=add_token_to_header(self, self.user, self.token)
```
Trinovantes: Was this test actually broken before?

Trinovantes:
Does this work only work for `GET` requests? i.e. Can `PUT` an entire Procedure JSON object with Page details?

src-django/api/tests/test_procedures.py
Ommy: done

src-django/api/tests/test_procedures.py
Ommy: When I removed the decorator from the method and added it to the `setUp` method, the call to `add_token_to_header` was throwing an error saying it needed a 3rd argument, where it's required params were `(self, user, token)`.

Ommy: And yeah by broken it never ran cause of the decorator, which I'm still confused as to why the decorator prevented the test from being run or maybe just from being found by the testing framework

Ommy:
That is an entirely different topic than what this ticket is doing

src-django/api/tests/test_procedures.py
Ommy: Previous tests also used double quotes for the data passed into the request, so I'll leave it.

src-django/api/tests/test_procedures.py
Ommy: I can change the method signature and remove `self` as a parameter since it isn't used, but I'm not sure if that's not a 'Pythony' way to do things. I'll need a second opinion on that.

src-django/api/tests/test_procedures.py
Trinovantes: Change it

Trinovantes:
It's fine, I already got a work around for the nested objects for now. We should eventually allow nested saving in the future as an optimization to reduce stress on the server.

src-django/api/tests/test_procedures.py
Trinovantes: If it's not used, might as well remove it

Trinovantes:
+1

ConnorCimowsky:
:+1:

PR #263, opened 2015-12-12
Minor fixes

Trinovantes:
- Improved CSS id/class names
- Reverted adding `/api` to `API_BASE` in `config.js`
- Replaced logger with preprocessor


Ommy:
LGTM. Does the `strip-debug` just disable console logs and `debugger` statements based on environment?

Trinovantes:
It reads in the JS files as a stream and removes "console" and "debugger" strings when enabled (e.g. when not using `gulp debug`)

PR #270, opened 2015-12-16
Removed Bower

### LINKED ISSUE #258: Remove bower
Seems like Bootstrap is available on `npm` as well. This should simplify our gulp file a bit more

Note our css main file needs this line:
```
@import "node_modules/bootstrap/less/bootstrap";
```

Trinovantes:
We only needed bower to load bootstrap but it seems npm also has a copy
of bootstrap.

Closes #258


Ommy:
+1

Issue #259, opened 2015-12-07
Restructure dependency injection


Trinovantes:
It's quite troublesome right now to inject our dependencies via the constructors

e.g.
```
constructor: function(options = {}) {
        this.app     = options.app     || global.App;
        this.Helpers = options.Helpers || require('utils/helpers');
        Marionette.ItemView.prototype.constructor.call(this, options);
},

// then later we have to call "this.Helpers.whatever"
```

There's some node libraries that help with dependency injections when writing tests such as https://www.npmjs.com/package/rewire

PR #276, opened 2016-01-07
Refactored code and disabled front end tests

### LINKED ISSUE #255: Rename templates to match their 1:1 view
I don't even know why I called some templates like `procedures_navbar.hbs` but their 1:1 corresponding view is `procedures_navbar_view.js`. These templates should all be renamed to match the file names of their corresponding view (if any).

Trinovantes:
- Removed dependency injection (part of #259)
- Disable frontend tests (going to rewrite the tests this weekend based
  on the new names)
- Replaced all instances of 'var' with 'let' (main difference is that
  var is local to the function whereas let is local to the block - which
  is what other traditional languages follow)
- Renamed everything from snake case to camel case where applicable
- Renamed templates to match their corresponding views

Closes #255


fabfile.py
```diff
@@ -17,7 +17,8 @@ def test():
         local('python manage.py test authentication --noinput')
 
     with lcd(env.frontend_dir):
-        local('npm test')
+        pass
```
Ommy: why?

PR #280, opened 2016-01-10
Fixed front end tests

### LINKED ISSUE #259: Restructure dependency injection

Trinovantes:
Misc:
- Renamed tests to camel case
- Replaced var with let
- Run tests with gulp instead of command line for more control

Changes:
- Use proxyquire for dependency injection for tests
- Moved SanaApp tests to new file
- Smoke test is literally a smoke test

Closes #259


Ommy:
lgtm

src-backbone/gulpfile.js
```diff
@@ -79,56 +79,57 @@ gulp.task('css', function() {
...
-    var filter = require('gulp-filter');
-
-    var js_filter = filter('**/*.js');
+    let jshint = require('gulp-jshint'); // Note: we're using JSHint, a better fork of the original JSLint
+    let stylish = require('jshint-stylish');
+    let filter = require('gulp-filter');
 
     gulp.src(Config.javascripts)
-        .pipe(js_filter)
+        .pipe(filter('**/*.js'))
```
Kapin: why did this change?

Kapin:
Just one question otherwise lgtm

src-backbone/gulpfile.js
Trinovantes: Wanted to clean up simple variables. In hindsight, this was a bad idea as I need to add this back because the filter needs more rules in another commit

Kapin: sounds good. lgtm after that

PR #281, opened 2016-01-10
Generate deployment-based config file

### LINKED ISSUE #252: Allow per-deployment JS config file
Right now all of our configuration is in `utils/config.js` file but we may want per-deployment specific configuration. For example on our production servers, we would want to set `DEBUG = true` and `API_BASE = {production API server}`.

Maybe `config.js` should export its config object merged with values found in `utils/config_overrides.js` (this file will be gitignored so we can set production server values).

Trinovantes:
Closes #252


Trinovantes:
`gulp build` in production server will generate a non-debug config file and `gulp debug` will generate a debug config file

Kapin:
:+1:

PR #282, opened 2016-01-10
Removed generating Django log file for auth module

### LINKED ISSUE #254: Do not let Django generate log file when in debug mode
Getting annoyed that Django is generating a `info.log` in my root directory when debugging.

Trinovantes:
I don't think we even read the info.log file that gets generated and it's quite annoying to see log files in my directories.

Closes #254


Ommy:
+1

ConnorCimowsky:
+1

PR #290, opened 2016-01-12
Removed decorator for tests

### LINKED ISSUE #261: Tests with initialize_permissions decorator don't run
The tests that have the `@initialize_permissions` decorator don't seem to be run when calling `./manage.py test`. I got around it by putting the method call to `#initialize_permissions` within the `#setUp` method. Maybe there's a nicer way, but we need to get this fixed up for the Pages tests.

Ommy:
The decorator was causing problems with our tests, and basically
preventing them from running. Now we call the grant_permissions method
directly.

Resolves #261


Kapin:
:+1:

src-django/api/tests/test_pages.py
```diff
@@ -27,9 +27,9 @@ def setUp(self):
             procedure=procedure
         )
         self.token = Token.objects.get(user=self.user)
-        self.partial_bulk_url = '/api/pages/partial_bulk_update/'
+        self.partial_bulk_url = '/api/pages/partial_bulk_update'
```
Trinovantes: should capitalize this to denote it as const

Trinovantes:
+1

PR #297, opened 2016-01-20
Add model and tests for concept

ConnorCimowsky:
~ paired with @Ommy ~

### DIFF
--- a/src-django/api/models.py
+++ b/src-django/api/models.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.contrib.auth.models import User
 from django.db import IntegrityError
+import uuid
 
 
 class Procedure(models.Model):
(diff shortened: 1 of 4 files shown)


src-django/api/tests/test_model_concept.py
```diff
@@ -0,0 +1,68 @@
...
+from django.db import IntegrityError
+from nose.tools import raises, assert_equals, assert_not_equals, assert_is_not_none
+from api.models import Concept
+from utils import factories
+
+
+class ConceptTest(TestCase):
+    def test_concept_required(self):
+        concept = Concept.objects.create(
+            name='SX SITE SWELLING',
```
Kapin: can you make this a variable

src-django/api/tests/test_model_concept.py
```diff
@@ -0,0 +1,68 @@
...
+            display_name='Swelling at surgical site'
```
Kapin: same here

Kapin:
2 nits otherwise lgtm

src-django/api/models.py
```diff
@@ -68,3 +69,30 @@ def save(self, **kwargs):
     class Meta:
         app_label = 'api'
         ordering = ['page', 'display_index']
+
+class Concept(models.Model):
+    TYPES = (
+        ('string', 'string'),
```
Trinovantes: why is this a pair?

ConnorCimowsky: because it was in Element ¯\_(ツ)_/¯

Trinovantes:
This doesn't look it integrates with Element?

ConnorCimowsky:
It does not, but we were following the convention we saw elsewhere. Do you know why it was done that way for Element?

Ommy:
resolves #283

PR #299, opened 2016-01-22
Create models for conditional logic

Kapin:
I'm not sure the best way to get the children # validation working because django doesn't allow for OneToMany as a specifier. Otherwise, this should be the basis for conditonals

### DIFF
--- a/src-django/api/models.py
+++ b/src-django/api/models.py
@@ -97,3 +97,78 @@ class Concept(models.Model):
 
     class Meta:
         app_label = 'api'
+
+
+class ConditionNode(models.Model):
+    CRITERIA_TYPES = (
+        ('EQUALS', 'EQUALS'),
+        ('GREATER', 'GREATER'),
+        ('LESSER', 'LESSER')
+    )
+
+    NODE_TYPES = (
+        ('AND', 'AND'),
+        ('OR', 'OR'),
+        ('NOT', 'NOT'),
+        ('CRITERIA', 'CRITERIA')
+    )
+
+    parent = models.ForeignKey('self', on_delete=models.CASCADE, related_name='children', blank=True, null=True)
+    criteria_element = models.ForeignKey(Element, blank=True, null=True)
+    node_type = models.CharField(max_length=8, choices=NODE_TYPES)
+    criteria_type = models.CharField(max_length=8, choices=CRITERIA_TYPES, null=True, blank=True)
+    value = models.TextField(blank=True, null=True)
+    last_modified = models.DateTimeField(auto_now=True)
+    created = models.DateTimeField(auto_now_add=True)
+
+    def save(self, **kwargs):
+        if (self.node_type, self.node_type) not in self.NODE_TYPES:
+            raise IntegrityError('Invalid condition node type')
+
+        if self.node_type is 'CRITERIA':
+            if not self.criteria_type:
+                raise IntegrityError('"CRITERIA" node type requires a criteria type')
+
+            if (self.criteria_type, self.criteria_type) not in self.CRITERIA_TYPES:
+                raise IntegrityError('Invalid criteria_type for condition')
+
+            # if self.parent.count() is not 0:
+            #     raise IntegrityError('CRITERIA can not have child nodes')
+
+            if not self.criteria_element:
+                raise IntegrityError('CRITERIA must have an element')
+        else:
+            if self.value:
+                raise IntegrityError('Only "CRITERIA" should have a value')
+
+            if self.criteria_element:
+                raise IntegrityError('Only "CRITERIA" should have an element')
+
+            if self.criteria_type:
+                raise IntegrityError('Only "CRITERIA" should have a criteria type')
+
+        # if self.node_type is 'NOT' and len(self.children) is not 1:
+        #     raise IntegrityError('NOT must have exactly one child node')
+
+        # if self.children.count < 2:
+        #     raise IntegrityError('AND and OR nodes must have at least two children')
+
+        super(ConditionNode, self).save()
+
+        if self.parent:
+            self.parent.last_modified = self.last_modified
+            self.parent.save()
+        elif self.show_if:
+            for s in self.show_if.all():
+                s.last_modified = self.last_modified
+                s.save()
+
+    class Meta:
+        app_label = 'api'
+
+
+class ShowIf(models.Model):
+    page = models.ForeignKey(Page)
+    condition = models.ForeignKey(ConditionNode, related_name='show_if')
+    last_modified = models.DateTimeField(auto_now=True)
+    created = models.DateTimeField(auto_now_add=True)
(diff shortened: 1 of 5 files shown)


src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
...
+
+    NODE_TYPES = (
+        ('AND', 'AND'),
+        ('OR', 'OR'),
+        ('NOT', 'NOT'),
+        ('CRITERIA', 'CRITERIA')
+    )
+
+    parent = models.ForeignKey('self', on_delete=models.CASCADE, related_name='children', blank=True, null=True)
+    element = models.ForeignKey(Element, blank=True, null=True)
```
connorcimowsky: should this be nullable?

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
...
+                raise IntegrityError('Invalid condition node type')
+
+            if self.node_type is 'CRITERIA':
+                if not self.criteria_type:
+                    raise IntegrityError('"CRITERIA" node type requires a criteria type')
+
+                if (self.criteria_type, self.criteria_type) not in self.CRITERIA_TYPES:
+                    raise IntegrityError('Invalid criteria_type for condition')
+
+                # if self.parent.count() is not 0:
```
connorcimowsky: I'm assuming these commented lines are what you meant about children # validation?

src-django/api/models.py
Trinovantes: A page has conditions, not element. Also a page should have the foreign key to condition, not the other way around

connorcimowsky: (not sure if you're replying to my comment, or making an unrelated comment about this line :stuck_out_tongue:)

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
...
+    parent = models.ForeignKey('self', on_delete=models.CASCADE, related_name='children', blank=True, null=True)
+    element = models.ForeignKey(Element, blank=True, null=True)
+    node_type = models.CharField(max_length=8, choices=NODE_TYPES, null=True, blank=True)
+    criteria_type = models.CharField(max_length=8, choices=CRITERIA_TYPES, null=True, blank=True)
+    value = models.TextField(blank=True, null=True)
+    last_modified = models.DateTimeField(auto_now=True)
+    created = models.DateTimeField(auto_now_add=True)
+
+    def save(self, **kwargs):
+        if self.node_type:
```
Trinovantes: shouldn't this throw an error if `node_type` is null?

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
 
     class Meta:
         app_label = 'api'
+
+
+class Condition(models.Model):
+    CRITERIA_TYPES = (
```
Trinovantes: why is everything in pairs?

connorcimowsky: https://docs.djangoproject.com/en/1.9/ref/models/fields/#choices

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
 
     class Meta:
         app_label = 'api'
+
+
+class Condition(models.Model):
```
Trinovantes: may as well call this `ConditionNode`

src-django/api/tests/test_model_condition.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def test_creates_and_condition(self):
+        Condition.objects.create(
+            node_type='AND'
+        )
+
+        condition = Condition.objects.get(node_type='AND')
+
+        assert_equals(condition.node_type, 'AND')
+        assert_not_equals(condition.last_modified, None)
```
connorcimowsky: use `assert_is_not_none(condition.last_modified)`?

connorcimowsky: (same goes for all other instances of `assert_not_equals(foo, None)`

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
...
+
+            # if self.children.count < 2:
+            #     raise IntegrityError('AND and OR nodes must have at least two children')
+
+        super(Condition, self).save()
+
+        if self.parent:
+            self.parent.last_modified = self.last_modified
+            self.parent.save()
+        elif self.show_if:
```
Trinovantes: where is `Condition.show_if` ever set?

src-django/api/models.py
```diff
@@ -97,3 +97,70 @@ def save(self, **kwargs):
...
+                if (self.criteria_type, self.criteria_type) not in self.CRITERIA_TYPES:
+                    raise IntegrityError('Invalid criteria_type for condition')
+
+                # if self.parent.count() is not 0:
+                #     raise IntegrityError('CRITERIA can not have child nodes')
+
+                if not self.element:
+                    raise IntegrityError('CRITERIA must have an element')
+
+            # if self.node_type is 'NOT' and len(self.children) is not 1:
```
Trinovantes: should also perform some assertion checks to make sure `AND` `OR` `NOT` nodes do not have any unnecessary fields filled

src-django/api/models.py
Trinovantes: I love concurrency

Kapin: Condition != ShowIf. ShowIf is the wrapper around the condition syntax tree. A ConditionNode (if it is a criteria) has a reference to the element it is referring to. I forgot to add the page foreign key to the ShowIf, and will do that.

This also should be nullable, because not all conditions have this. We check to make sure the Criteria ones do though.

src-django/api/models.py
Kapin: Yes, I haven't figured out a good way to do this yet.

src-django/api/models.py
Kapin: If the front-end will guarentee that the node_type is sent up, yes. Need to discuss how the front end works when it comes to these validations in general in person.

src-django/api/tests/test_model_condition.py
Kapin: I'll do a sweep of those.

src-django/api/models.py
Kapin: Makes sense

src-django/api/models.py
Kapin: It's referenced from the ShowIf model as a foreign key to a condition. There is no such thing as a OneToMany relationship, so we have to do it this way. It's kinda silly...

src-django/api/models.py
Kapin: Can do that. The original idea was to ignore them.

src-django/api/models.py
Trinovantes: it may be clearer to call it `operand_element`

src-django/api/models.py
Trinovantes: remember that we are exposing the API to the public so it's always good to have as much sanity checks as possible

Trinovantes:
+1

Ommy:
lgtm

PR #304, opened 2016-02-11
Implement condition logic serialization

Kapin:
Still missing a lot of tests, but this should be the serialization and api for condition logic.

Example POST:

```
{
    "page":1,
    "conditions": [
        {
            "node_type": "NOT",
            "children": [
                {
                    "node_type": "AND",
                    "children": [
                        {
                            "criteria_element": 3,
                            "node_type": "CRITERIA",
                            "criteria_type": "EQUALS",
                            "value": "foo"
                        },
                        {
                            "criteria_element": 3,
                            "node_type": "CRITERIA",
                            "criteria_type": "EQUALS",
                            "value": "bar"
                        }
                    ]
                }
            ]
        }
    ]
}
```

### DIFF
--- a/src-django/api/models.py
+++ b/src-django/api/models.py
@@ -99,6 +99,18 @@ class Concept(models.Model):
         app_label = 'api'
 
 
+class ShowIf(models.Model):
+    page = models.ForeignKey(Page, related_name='show_if')
+    last_modified = models.DateTimeField(auto_now=True)
+    created = models.DateTimeField(auto_now_add=True)
+
+    def save(self, **kwargs):
+        super(ShowIf, self).save()
+
+        self.page.last_modified = self.last_modified
+        self.page.save()
+
+
 class ConditionNode(models.Model):
     CRITERIA_TYPES = (
         ('EQUALS', 'EQUALS'),
@@ -114,6 +126,7 @@ class ConditionNode(models.Model):
     )
 
     parent = models.ForeignKey('self', on_delete=models.CASCADE, related_name='children', blank=True, null=True)
+    show_if = models.ForeignKey(ShowIf, on_delete=models.CASCADE, related_name='conditions', blank=True, null=True)
     criteria_element = models.ForeignKey(Element, blank=True, null=True)
     node_type = models.CharField(max_length=8, choices=NODE_TYPES)
     criteria_type = models.CharField(max_length=8, choices=CRITERIA_TYPES, null=True, blank=True)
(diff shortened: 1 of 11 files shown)


src-django/api/views.py
```diff
@@ -80,3 +80,12 @@ class ConceptViewSet(viewsets.ModelViewSet):
...
         return models.Concept.objects
+
+
+class ShowIfViewSet(viewsets.ModelViewSet):
+    model = models.ShowIf
+    serializer_class = serializer.ShowIfSerializer
+
+    def get_queryset(self):
+        user = self.request.user
+        return models.ShowIf.objects.filter(page__procedure__owner_id__exact=user.id)
```
Trinovantes: One thing I noticed is that we don't validate the `self.request.user` object in our viewsets so we keep getting 500 (accessing `id` of `null`) instead of 401

Trinovantes:
Should clarify documentation a bit more

+1

src-django/api/views.py
Kapin: separate pr, but yes good idea.

Kapin:
@Trinovantes besides example queries, any suggestions?

Ommy:
Overall looks great but I don't see any validation for the `NOT` type.

connorcimowsky:
LGTM; where do you mean, @Ommy?

Kapin:
@Ommy the validation for NOT would be that is has exactly one child. I need to go in and add a bunch of validations around that, it is just a teeny bit challenging.

src-django/api/models.py
```diff
@@ -147,6 +161,12 @@ def save(self, **kwargs):
             if self.criteria_type:
                 raise IntegrityError('Only "CRITERIA" should have a criteria type')
 
+        if self.show_if and self.parent:
+            raise IntegrityError('Can not have both a show_if and parent')
+
+        if not self.show_if and not self.parent:
+            raise IntegrityError('Condition node must have a parent or show_if')
```
Trinovantes: Comment should be "xor" instead of "or"

Trinovantes:
I looked through the examples in https://github.com/heywbj/django-rest-framework-recursive so it should be fine.

src-django/api/models.py
Kapin: sure

Ommy:
Yeah I meant children validation for NOT. So @Kapin, is that validation coming in another PR?

Kapin:
@Ommy yes

PR #305, opened 2016-02-16
Make `concept` a foreign key on `Element`

### LINKED ISSUE #283: Create Concept model, serializer, generator on backend

connorcimowsky:
Resolves #283.

Used to be a `TextField`.

### DIFF
--- a/src-django/api/models.py
+++ b/src-django/api/models.py
@@ -78,7 +78,7 @@ class Element(models.Model):
     element_type = models.CharField(max_length=12, choices=TYPES, null=True, blank=True)
     choices = models.TextField(null=True, blank=True)
     numeric = models.CharField(max_length=255, null=True, blank=True)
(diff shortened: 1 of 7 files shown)


src-django/api/models.py
```diff
@@ -34,6 +34,34 @@ class Meta:
...
 
+class Concept(models.Model):
+    TYPES = (
+        ('string', 'string'),
+        ('boolean', 'boolean'),
+        ('number', 'number'),
+        ('complex', 'complex')
+    )
+
+    uuid = models.UUIDField(default=uuid.uuid4, null=False, blank=False, editable=False)
```
Kapin: can we enforce proper constraints here? Everything that is blankable/nullable should be and everything that isn't, shouldn't.

Kapin:
LGTM

PR #309, opened 2016-02-17
Add a ton of validations to the serializers

Kapin:


src-django/api/serializer.py
```diff
@@ -48,6 +49,13 @@ def to_representation(self, instance):
 
         return ret
 
+    def validate(self, data):
+        if 'element_type' in data and data['element_type'] in models.Element.CHOICE_TYPES:
```
Trinovantes: `element_type` attribute should be enforced on the server

Trinovantes: also if it's not in `CHOICE_TYPES`, it should also throw an error

src-django/api/serializer.py
```diff
@@ -48,6 +49,13 @@ def to_representation(self, instance):
...
+            if 'answer' in data and data['answer'] and data['answer'] not in data['choices']:
```
Trinovantes: this validation should not be indented

src-django/api/serializer.py
```diff
@@ -98,6 +108,39 @@ def create(self, validated_data):
...
+        if data['node_type'] in models.ConditionNode.CRITERIA_TYPES:
+            if 'children' in data:
+                raise serializers.ValidationError('CRITERIA node must have no children')
+
+            if 'criteria_element' not in data:
+                raise serializers.ValidationError('CRITERIA must have an element')
+
+            if 'value' not in data:
+                raise serializers.ValidationError('CRITERIA must have a value')
+        else:
```
Trinovantes: make a comment that `node_type` is guaranteed to be one of `NODE_TYPES`

src-django/api/serializer.py
```diff
@@ -98,6 +108,39 @@ def create(self, validated_data):
...
+            if 'value' in data:
+                raise serializers.ValidationError('Only "CRITERIA" should have a value')
+
+            if 'criteria_element' in data:
+                raise serializers.ValidationError('Only "CRITERIA" should have an element')
+
+        if data['node_type'] == 'NOT' and 'children' in data and len(data['children']) > 1:
```
Trinovantes: this and the next if statement could be indented because they are logical types

Trinovantes:
+1

PR #312, opened 2016-02-18
Fix bugs relating to validation

Kapin:
Because I'm bad at reading my terminal


Trinovantes:
Add a comment warning what we can't validate

PR #314, opened 2016-02-18
Moved session fetch to init() method

Trinovantes:
This should fix the timeouts some tests are getting due to calling
_setupSession() with fetch() inside.


Kapin:
it better. +1

PR #315, opened 2016-02-20
Add functionality for sending and verifying email activation keys

connorcimowsky:
To be addressed in the future:

- Allow resending of activation email
- Send new key when email is changed
- HTML templates with i18n support


Ommy:
Instead of storing expiring email tokens in the database, which will just take up space after expired, why not use redis? We can set the key value pair to expire after the 2 days and we won't need to have a separate task that goes through the database and delete old expired tokens every week.



From [redis](https://pypi.python.org/pypi/redis) can do something like `redis.setex(key, expiry, value)` where expiry is in seconds.

Ommy:
Also, for your last point about HTML templates with i18n support for emails, I also needed to get that done for my ticket. I made an issue for that as well, which is #317

src-django/authentication/tests/utils/factories.py
```diff
@@ -0,0 +1,23 @@
...
+    is_email_confirmed = False
+
+
+class EmailConfirmationKeyFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = EmailConfirmationKey
+        django_get_or_create = ('user',)
+
+    user = factory.SubFactory(factories.UserFactory)
+    key = '84bff1c64351a616828596f01b4e0d3155190eb5'
```
Trinovantes: How is this key being used? If it's to be used in a HMAC function, maybe we should use the secrete Django key?

src-django/authentication/views.py
```diff
@@ -114,3 +126,26 @@ def logout(request):
...
+    if email_confirmation_key.expiration < timezone.now():
+        return HttpResponseBadRequest("Key expired")
+
+    user_profile = email_confirmation_key.user.profile
+    user_profile.is_email_confirmed = True
+    user_profile.save()
+
+    email_confirmation_key.delete()
+
+    return JsonResponse({
```
Trinovantes: Does this need a front end?

src-django/authentication/tests/utils/factories.py
connorcimowsky: This particular string constant, or the email confirmation key in general?

src-django/authentication/views.py
connorcimowsky: Yes, ideally :)

src-django/authentication/tests/utils/factories.py
Trinovantes: This constant. And how is the email confirmation key being generated?

connorcimowsky: This is just a factory for tests. The generation code is in [src-django/authentication/models.py](https://github.com/SanaMobile/sana.protocol_builder/blob/connor/activation-redux/src-django/authentication/models.py).

src-django/authentication/models.py
```diff
@@ -1 +1,32 @@
-from signals import user_post_save  # noqa
+from django.db import models
+from django.contrib.auth.models import User
+import datetime
+import hashlib
+import random
+
+EMAIL_CONFIRMATION_KEY_TTL_DAYS = 2
```
Kapin: This might be better inside the EmailConfirmationKey class

connorcimowsky: Good call. :sparkles:

src-django/authentication/models.py
```diff
@@ -1 +1,32 @@
...
+    user = models.OneToOneField(User, related_name="profile", on_delete=models.CASCADE)
+    is_email_confirmed = models.BooleanField(default=False)
+
+    class Meta:
+        app_label = 'authentication'
+
+
+class EmailConfirmationKey(models.Model):
+    user = models.OneToOneField(User)
+    key = models.CharField(max_length=40, null=False, blank=False)
```
Kapin: You don't need to explicitly say that these aren't false or blank. It is by default.

src-django/authentication/tests/test_confirm_email.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ConfirmEmailTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_valid_key(self):
+        email_confirmation_key = factories.EmailConfirmationKeyFactory()
+        response = self.client.get('/auth/confirm_email/{0}'.format(email_confirmation_key.key))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_confirm_email.py
```diff
@@ -0,0 +1,53 @@
...
+        assert_equals(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_user_becomes_confirmed(self):
+        user_profile = factories.UserProfileFactory()
+        email_confirmation_key = factories.EmailConfirmationKeyFactory(user=user_profile.user)
+
+        assert_false(user_profile.is_email_confirmed)
+
+        response = self.client.get('/auth/confirm_email/{0}'.format(email_confirmation_key.key))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_confirm_email.py
```diff
@@ -0,0 +1,53 @@
...
+
+        user_profile.refresh_from_db()
+        assert_true(user_profile.is_email_confirmed)
+
+    def test_key_is_deleted(self):
+        email_confirmation_key = factories.EmailConfirmationKeyFactory()
+        response = self.client.get('/auth/confirm_email/{0}'.format(email_confirmation_key.key))
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_model_email_confirmation_key.py
```diff
@@ -0,0 +1,17 @@
+from api.tests.utils.factories import UserFactory
+from authentication.models import EmailConfirmationKey
+from django.test import TestCase
+from freezegun import freeze_time
+from nose.tools import assert_equals
+import datetime
+
+
+class EmailConfirmationKeyTest(TestCase):
+    @freeze_time("2016-02-19")
```
Kapin: I like this :smile:

src-django/authentication/tests/test_signup.py
```diff
@@ -17,10 +19,36 @@ def test_new_user_can_signup(self):
...
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
         r = json.loads(response.content)
         assert_true(r['success'])
         assert_true(r['token'] is not None)
         assert_equals(len(r['errors']), 0)
+
+    def test_new_user_is_unconfirmed(self):
+        response = self.client.post('/auth/signup', self.valid_fields)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_signup.py
```diff
@@ -17,10 +19,36 @@ def test_new_user_can_signup(self):
...
+
+        user = User.objects.get(username='admin')
+        assert_false(user.profile.is_email_confirmed)
+
+    def test_email_confirmation_key_is_created(self):
+        response = self.client.post('/auth/signup', self.valid_fields)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_signup.py
```diff
@@ -17,10 +19,36 @@ def test_new_user_can_signup(self):
...
+        user = User.objects.get(username='admin')
+        email_confirmation_key = EmailConfirmationKey.objects.get(user=user)
+        assert_is_not_none(email_confirmation_key)
+
+    @override_settings(CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
+                       CELERY_ALWAYS_EAGER=True,
+                       BROKER_BACKEND='memory')
+    def test_email_is_sent(self):
+        response = self.client.post('/auth/signup', self.valid_fields)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
Kapin: assert_equals

src-django/authentication/tests/test_confirm_email.py
connorcimowsky: oh weird, I was following the convention in [src-django/authentication/tests/test_signup.py](https://github.com/SanaMobile/sana.protocol_builder/blob/master/src-django/authentication/tests/test_signup.py), but I agree it should be `assert_equals`

Kapin: We should fix that

src-django/authentication/tests/test_signup.py
```diff
@@ -17,10 +19,36 @@ def test_new_user_can_signup(self):
...
         assert_equals(len(r['errors']), 0)
+
+    def test_new_user_is_unconfirmed(self):
+        response = self.client.post('/auth/signup', self.valid_fields)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        user = User.objects.get(username='admin')
+        assert_false(user.profile.is_email_confirmed)
+
+    def test_email_confirmation_key_is_created(self):
```
Ommy: Since you're doing a valid `POST` here as well, maybe you want to set the `BROKER_BACKEND` to `memory`? I think that's why the tests are failing.

Kapin:
+1

Trinovantes:
+1

connorcimowsky:
:rotating_light: redis implementation is in, please review :rotating_light: https://github.com/SanaMobile/sana.protocol_builder/commit/4523bee055af2837ad49e52ba8e9e00e28d87b2c

Kapin:
LGTM. Make sure someone else looks too.

src-django/authentication/views.py
```diff
@@ -114,3 +141,28 @@ def logout(request):
...
+# Confirm Email
+# ------------------------------------------------------------------------------
+
+@csrf_exempt
+@require_http_methods(['GET'])
+def confirm_email(request, key):
+    user_id = cache.get(key)
+
+    if user_id is None:
+        return HttpResponseNotFound("Invalid or expired key")
```
Trinovantes: is there a way to return a JSON response with 404 status code?

connorcimowsky: done

connorcimowsky:
(also switched over to a html template for the email)

src-django/authentication/views.py
```diff
@@ -50,6 +68,16 @@ def signup(request):
         user = form.save()
         token_key = Token.objects.get(user=user).key
         logger.info("Signup Success: u:{0} e:{0}".format(user.username, user.email))
+
+        key = _generateEmailConfirmationKey(user.email)
+        timeout = _emailConfirmationKeyTimeout()
+        cache.set(key, user.pk, timeout)
+
+        # TODO(connor): remove hard-coded url
```
Ommy: Might as well make an issue

Ommy:
+1

PR #320, opened 2016-02-23
Added a redis cache system for django

### LINKED ISSUE #316: Sana should have a redis client
We already have a redis client set up with Chef for Celery, but we need to have a redis class in Sana that establishes a connection to redis and gives us the client object.

We can just use https://pypi.python.org/pypi/redis, giving it configuration from `settings.py`

We also need to have this class give us a "fake" redis client for testing purposes, just so we don't need to start up a redis client when running builds on travis. We can use https://github.com/jamesls/fakeredis

Ommy:
This allows us to connect to redis using the internal django cache libraries. Redis is already running for Celery, this allows us to connect to it via django's internal caching system. We can use:

`from django.core.cache import cache` and then use redis library functions like `set` and `get`

And then, when we want to test functionality, we can use django's `@override_settings` or `@modify_settings`

```
@modify_settings(CACHES={
        'default': 'django.core.cache.backends.locmem.LocMemCache',
})
```

Resolves  #316

### DIFF
--- a/src-django/requirements.txt
+++ b/src-django/requirements.txt
@@ -7,6 +7,7 @@ django-compressor==2.0
 django-cors-headers==1.1.0
 django-filter==0.12.0
 django-nose==1.4.3
+django-redis-cache==1.6.5
 djangorestframework==3.3.2
 djangorestframework-recursive==0.1.1
 django-extensions==1.6.1
@@ -16,3 +17,4 @@ gunicorn==19.4.5
 nose==1.3.7
 psycopg2==2.6.1
 redis==2.10.3
+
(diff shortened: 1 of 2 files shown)


Trinovantes:
What are some possible use cases for this cache?

Ommy:
Well main use case would be for temporary token storage such as, and this is if @connorcimowsky decides to go this route, email conformation tokens and password reset tokens. We already have redis running for Celery, this just allows us to set Django's internal cache mechanism to use that redis.

src-django/requirements.txt
```diff
@@ -16,3 +16,4 @@ gunicorn==19.4.5
 nose==1.3.7
 psycopg2==2.6.1
 redis==2.10.3
+django-redis-cache==1.6.5
```
Kapin: alphabetize plz

Kapin:
LGTM

connorcimowsky:
ship ittttt

PR #322, opened 2016-02-23
Use debounce instead of delaySave for proper request throttling

Kapin:
This should properly throttle updating while typing in any field.

### DIFF
--- a/src-backbone/app/js/views/builder/builderHeaderView.js
+++ b/src-backbone/app/js/views/builder/builderHeaderView.js
@@ -19,9 +19,7 @@ module.exports = Marionette.ItemView.extend({
         'click @ui.downloadButton': '_download'
     },
 
-    _save: function() {
-        Helpers.delaySave(this, this._saveToServer);
-    },
+    _save: _.debounce(this._saveToServer(), Config.INPUT_DELAY_BEFORE_SAVE),
 
     _saveToServer: function() {
         this.model.save({
--- a/src-backbone/app/js/sanaApp.js
+++ b/src-backbone/app/js/sanaApp.js
@@ -128,7 +128,7 @@ module.exports = Marionette.Application.extend({
     _setupBackboneSync: function() {
         let self = this;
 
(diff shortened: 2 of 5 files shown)


src-backbone/app/js/views/builder/builderHeaderView.js
```diff
@@ -19,9 +19,9 @@ module.exports = Marionette.ItemView.extend({
         'click @ui.downloadButton': '_download'
     },
 
-    _save: function() {
-        Helpers.delaySave(this, this._saveToServer);
-    },
+    _save: _.debounce(function(event) {
```
Trinovantes: this could be simplified to just `_.debounce(this.saveToServer, Config.INPUT_DELAY_BEFORE_SAVE)`

src-backbone/app/js/views/builder/builderHeaderView.js
```diff
@@ -19,9 +19,9 @@ module.exports = Marionette.ItemView.extend({
...
+        this._saveToServer();
+    }, 1000),
```
Trinovantes: use the config constant instead of hardcoding

src-backbone/app/js/models/session.js
```diff
@@ -126,7 +126,7 @@ let SessionModel = Backbone.Model.extend({
             },
             error: function(errors) {
                 App().RootView.clearNotifications();
-                let parsedErrors = JSON.parse(errors.responseText)
+                let parsedErrors = JSON.parse(errors.responseText);
```
Trinovantes: Leave this warning. I'm going to fix this and the account settings UI altogether later

Trinovantes:
I'm not sure what bug this fixes but it does clean up the code

+1

Kapin:
@Trinovantes I'd rather go with a standardized implementation, than writing something that is hard to understand and hasn't been as tested :stuck_out_tongue:

Ommy:
+1

PR #324, opened 2016-02-23
Added jinja2 template loader module & template dir

### LINKED ISSUE #317: Sana should have some way of storing and using email templates in the backend
Right now we need email templates for email confirmation and password reset, and may have more in the future. We should set up some way of creating emails through custom templates. Django has some documentation around templates here: https://docs.djangoproject.com/en/1.9/topics/templates/. Alternatively, we can look into https://github.com/artemrizhov/django-mail-templated

Ommy:
This commit adds a template directory under `api` for storing our jinja templates. The templater module is just to load the template environment. Then, once you have the environment object returned by `get_environment` you can:

Load up your template by name: `template = env.get_template('template_name.html')`
And then you can render it: `template.render(your='variables', go='here')`

So, if we have a template like
```
Hello {{ name }},  I'm {{ sender }}.
```

we can render the template like this: `template.render(name='Geoff', sender='Lenny')`

and we'll get `Hello Geoff, I'm Lenny`

Resolves #317

### DIFF
--- a/src-django/mailer/templater.py
+++ b/src-django/mailer/templater.py
@@ -1,0 +1,5 @@
+from jinja2 import Environment, PackageLoader
+
+
+def get_environment(application, template_dir):
+    return Environment(loader=PackageLoader(application, template_dir))


Kapin:
lgtm

Trinovantes:
Don't you need an `__init__.py` inside the template folder?

Ommy:
No, the template loader will handle getting the files from that directory. There aren't going to be any `.py` files in `templates/` either.

Trinovantes:
+1

PR #346, opened 2016-02-24
Send back user info on login

### LINKED ISSUE #337: /login should return basic User details
Such as username and email so that they can be stored in the global `App.session` object and retrieved for the navbar

Ommy:
Includes if user is super user to use to determine if they can upload
a concept dictionary CSV.

Resolves #337

### DIFF
--- a/src-django/authentication/views.py
+++ b/src-django/authentication/views.py
@@ -99,6 +99,7 @@ def login(request):
     form = AuthenticationForm(data=request.POST)
     valid_form = form.is_valid()
     token_key = None
+    user_details = {}
 
     if valid_form:
         form.clean()  # Calls authenticate(); will through ValidationException if form is invalid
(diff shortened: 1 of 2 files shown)


src-django/authentication/views.py
```diff
@@ -114,6 +115,14 @@ def login(request):
 
         token_key = token.key
         logger.info("Login Success: u:{0} e:{0}".format(user.username, user.email))
+
+        user_details = {
+            'super_user': user.is_superuser,
```
Trinovantes: keep the keys the same

src-django/authentication/views.py
```diff
@@ -121,6 +130,7 @@ def login(request):
         'success': valid_form,
         'errors': form.errors,
         'token': token_key,
+        'user': user_details if user_details else False,
```
Trinovantes: this should return an empty object

src-django/authentication/views.py
```diff
@@ -99,6 +99,7 @@ def login(request):
     form = AuthenticationForm(data=request.POST)
     valid_form = form.is_valid()
     token_key = None
+    user_details = None
```
Trinovantes: you could init this to `{}` so the return statement doesn't need to check for None

Trinovantes:
+1

Kapin:
:+1:

PR #347, opened 2016-02-25
Add CSV upload functionality to the backend

connorcimowsky:


src-django/api/views.py
```diff
@@ -123,6 +125,49 @@ class ConceptViewSet(viewsets.ModelViewSet):
     def get_queryset(self):
         return models.Concept.objects
 
+    @list_route(methods=['POST'])
+    def import_csv(self, request):
+        if 'csv' not in request.FILES:
+            return JsonResponse({
```
Ommy: status for the JSONResponse?

src-django/api/views.py
```diff
@@ -123,6 +125,49 @@ class ConceptViewSet(viewsets.ModelViewSet):
...
+                    display_name="display_name",
+                    description="description",
+                    data_type="datatype",
+                    mime_type="mimetype",
+                    constraint="constraint"
+                )
+            )
+            copy_mapping.save()
+        except Exception as e:
+            return JsonResponse({
```
Ommy: status here too?

Ommy:
LGTM after comments

src-django/api/views.py
connorcimowsky: done for both.

src-django/api/views.py
```diff
@@ -123,6 +125,55 @@ class ConceptViewSet(viewsets.ModelViewSet):
...
+
+        with open(file_path, 'wb+') as destination:
+            for chunk in csv.chunks():
+                destination.write(chunk)
+
+        try:
+            copy_mapping = CopyMapping(
+                models.Concept,
+                file_path,
+                dict(
```
Kapin: Can this use the {} syntax or any specific reason that dict() is used?

Kapin: Also this may be better as a constant somewhere

src-django/api/views.py
```diff
@@ -123,6 +125,55 @@ class ConceptViewSet(viewsets.ModelViewSet):
...
+                    display_name="display_name",
+                    description="description",
+                    data_type="datatype",
+                    mime_type="mimetype",
+                    constraint="constraint"
+                )
+            )
+            copy_mapping.save()
+        except Exception as e:
+            return JsonResponse(
```
Kapin: Assuming this will be removed later? This leaks debug info.

src-django/api/views.py
connorcimowsky: No reason, just copied example code. Changed to `{}` and moved into a constant.

src-django/api/views.py
connorcimowsky: Good point; what would you recommend instead?

Kapin: errors should be blank? What are the scenarios that it would throw an exception?

connorcimowsky: There are 2 scenarios:

1. There are invalid columns in the uploaded CSV (the most likely scenario)
2. There was some sort of SQL error (very unlikely)

I'll handle the two exceptions separately, giving a more useful error message (and 400 status code) for the first.

connorcimowsky: Addressed in https://github.com/SanaMobile/sana.protocol_builder/commit/b3dfb848362156906ae2a31e508806ef88342729

Kapin:
One nit, but LGTM. Don't need to see it again.

src-django/api/views.py
```diff
@@ -117,12 +120,64 @@ def get_queryset(self):
...
+                status=status.HTTP_400_BAD_REQUEST,
+                data={
+                    'success': False,
+                    'errors': ["Missing 'csv' parameter"]
+                }
+            )
+
+        csv = request.FILES['csv']
+
+        file_path = "/tmp/{0}.csv".format(str(uuid.uuid4()))
```
Trinovantes: shouldn't you check if the file exists or not first (even though it is statistically rare to happen)?

src-django/api/views.py
```diff
@@ -117,12 +120,64 @@ def get_queryset(self):
...
+            expected_columns = ', '.join(ConceptViewSet.CSV_COLUMN_MAPPING.keys())
+            return JsonResponse(
+                status=status.HTTP_400_BAD_REQUEST,
+                data={
+                    'success': False,
+                    'errors': ['CSV import error. Expected columns are {0}'.format(expected_columns)]
+                }
+            )
+
+        return JsonResponse({
```
Trinovantes: why not just delete the tmp file when you're done instead of relying on the OS?

src-django/api/views.py
Kapin: It's statistically impossible. It's a UUID.

Trinovantes: But still not 0

src-django/api/views.py
connorcimowsky: Thanks for pointing that out — forgot to do that.

Issue #318, opened 2016-02-21
Race condition when trying to update criteria 


Trinovantes:
If I try to update a criteria in 2 quick requests, there's a race condition that causes the first response to return 2 nodes and second response to return 0 nodes.

Request 1:
```
{
   "page":28,
   "id":60,
   "last_modified":"2016-02-21T23:26:14.286201Z",
   "created":"2016-02-21T23:26:14.252864Z",
   "conditions":[
      {
         "criteria_element":null,
         "node_type":"EQUALS",
         "value":"asd"
      }
   ]
}
```

Request 2:
```
{
   "page":28,
   "id":60,
   "last_modified":"2016-02-21T23:26:14.286201Z",
   "created":"2016-02-21T23:26:14.252864Z",
   "conditions":[
      {
         "criteria_element":null,
         "node_type":"EQUALS",
         "value":"asdf"
      }
   ]
}
```

Response 1:
```
{
   "id":60,
   "page":28,
   "last_modified":"2016-02-21T23:26:17.789627Z",
   "created":"2016-02-21T23:26:14.252864Z",
   "conditions":[
      {
         "id":172,
         "parent":null,
         "criteria_element":null,
         "node_type":"EQUALS",
         "value":"asd",
         "last_modified":"2016-02-21T23:26:17.778803Z",
         "created":"2016-02-21T23:26:17.778831Z",
         "show_if":60,
         "children":[

         ]
      },
      {
         "id":173,
         "parent":null,
         "criteria_element":null,
         "node_type":"EQUALS",
         "value":"asdf",
         "last_modified":"2016-02-21T23:26:17.779719Z",
         "created":"2016-02-21T23:26:17.779743Z",
         "show_if":60,
         "children":[

         ]
      }
   ]
}
```

Response 2:
```
{
   "id":60,
   "page":28,
   "last_modified":"2016-02-21T23:26:17.753356Z",
   "created":"2016-02-21T23:26:14.252864Z",
   "conditions":[

   ]
}
```

Kapin:
Looking into this. However, from a client-perspective, wouldn't it make sense to ensure that the first save went through before the second one?

Trinovantes:
I worked around this by adding a sleep between requests. I think it's safer to make sure the server can handle these cases than trusting the client. (e.g. maybe an MITM can perform a replay attack to make the user lose data?)

Kapin:
[...]

PR #329, opened 2016-02-24
Conditional Logic UI

Trinovantes:
Closes #321 

screenshot from 2016-02-24 01 00 41

PR #349, opened 2016-02-25
Convert conditional logic to be stored as a string versus tree of nodes

### LINKED ISSUE #328: Add a display order field in condition nodes
_Sometimes_ the server returns the nodes not in the same order as the request which causes the front-end to resort the user's conditions.

Kapin:
Changes from node based structure to storing as a big string. JSON stays the same but key differences are:

1. `parent` foreign key is removed
2. `show_if` foreign key is removed
3. no timestamps
4. `AND` and `OR` nodes must have at least 1 child when sent up
5. `NOT` nodes must have exactly 1 child
6. No ids for anything except the `show_if` object

Requests should not be any different otherwise.

Closes #328 and #318 (in theory)


Trinovantes:
As long as this doesn't break #329

Kapin:
@Trinovantes it will require that to be slightly changed (renaming and such). It's not major.

src-django/api/tests/utils/factories.py
```diff
@@ -89,41 +89,4 @@ class Meta:
...
-
-
-class NotConditionFactory(factory.django.DjangoModelFactory):
-    class Meta:
-        model = models.ConditionNode
-
-    parent = None
-    node_type = 'NOT'
-    show_if = factory.SubFactory(ShowIfFactory)
+    conditional_tree = ""
```
Ommy: No tree by default?

Kapin: Getting the element foreign keys to behave was annoying. It's mainly a convenience factory.

Trinovantes:
Keep the interface the same. That's one of the reasons we kept the front end and backend separate

Kapin:
@Trinovantes I'm not returning an array when we don't need one. I will rename the field on the model.

src-django/api/field.py
```diff
@@ -14,3 +15,56 @@ def to_internal_value(self, data):
...
+            raise serializers.ValidationError('Missing or invalid node type')
+
+        if data['node_type'] in ShowIf.CRITERIA_TYPES:
+            if 'children' in data:
+                raise serializers.ValidationError('CRITERIA node must have no children')
+
+            if 'criteria_element' not in data or not isinstance(data['criteria_element'], int):
+                raise serializers.ValidationError('CRITERIA must have an element')
+
+            try:
```
Trinovantes: This check should only be in the generator. A user should be allowed to create an empty criteria node on the front end

src-django/api/field.py
```diff
@@ -14,3 +15,56 @@ def to_internal_value(self, data):
...
+            if 'criteria_element' in data:
+                raise serializers.ValidationError('Only "CRITERIA" should have an element')
+
+            if 'children' not in data or not isinstance(data['children'], list):
+                raise serializers.ValidationError('Logical nodes must have children')
+
+            if data['node_type'] == 'NOT' and len(data['children']) != 1:
+                raise serializers.ValidationError('NOT nodes must have exactly 1 child')
+
+            if data['node_type'] in ('AND', 'OR') and len(data['children']) < 2:
```
Trinovantes: AND and OR nodes should be allowed to have 1 child since it makes no logical difference to the user. The generator should unroll any logical nodes with only 1 child

Ommy:
+1

Trinovantes:
+1

PR #352, opened 2016-02-28
Add route for getting elements before a specific one

Kapin:
Needed for #332 

Creates an endpoint that given an element, returns all elements that occur in pages before it in a procedure.


Trinovantes:
I'm trying to think of whether or not this is better than computing the elements on the frontend and save a network request. Don't merge this yet

Trinovantes:
This change isn't necessary anymore.

Issue #333, opened 2016-02-24
Auto-complete for value field in Criteria (if possible)


Trinovantes:
- If Element is choice-based and type is EQUALS
- If Element is DATE

Kapin:
This should just be a drop-down in this case.

PR #354, opened 2016-02-29
Conditional Enhancements

Trinovantes:
This will contain all the missing enhancements for conditional logic UI. I would suggest you read each commit individually (skip the second commit since that's just reorganization).

- [x] Highlight pages in sidebar containing operand elements
- [ ] Show preview of condition in page list (#331)
- [x] Auto-complete for elements (#332)
- [x] Auto-complete for value (if possible) (#333)
- [x] First page should have no options to create conditional logic
- [x] If page is moved to first page, condition should be removed
- [x] If page P is moved above a page that contains elements that P depends on, condition should be removed


Kapin:
@Trinovantes can we review/merge what is in here so far?

Trinovantes:
Sure?

Trinovantes:
I'm going to cherry pick the choices-related commits to your branch and fix the multi-answers at the same time. This PR will only contain changes related to conditionals

connorcimowsky:
As per Josh's suggestion, :+1:

connorcimowsky:
BTW, my work is based on this branch — it would be great if we could merge this soon

Kapin:
This is fine to me. Merge please before doing any more work.

PR #363, opened 2016-03-02
Upgrade requirements

Trinovantes:
This change allows the transition to 1.9 if we want to upgrade now or later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ install:
   - pip install --requirement src-django/requirements.txt
 
   # Node tools
+  - npm set progress=false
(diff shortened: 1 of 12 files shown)


Trinovantes:
This works so may as well merge it in now ¯\_(ツ)_/¯

Kapin:
Can you squash this into one commit in case we run into other problems

Trinovantes:
The only thing that we need to undo (if there's any problems) is the actual bump. The first commit is compatible with 1.8.x

Kapin:
lgtm then

PR #370, opened 2016-03-04
Frontend view for concept uploader

Ommy:
Only displays on right navbar when user is a superuser

### DIFF
--- a/src-backbone/app/js/controllers/conceptsController.js
+++ b/src-backbone/app/js/controllers/conceptsController.js
@@ -1,20 +1,20 @@
 const Config = require('utils/config');
 
-let App                  = require('utils/sanaAppInstance');
-let Helpers              = require('utils/helpers');
-let ConceptsLayout       = require('views/concepts/conceptsLayoutView');
-let ConceptsUploaderView = require('views/concepts/conceptsUploaderView');
-
+const App                  = require('utils/sanaAppInstance');
+const Helpers              = require('utils/helpers');
+const ConceptsLayout       = require('views/concepts/conceptsLayoutView');
+const ConceptsUploaderView = require('views/concepts/conceptsUploaderView');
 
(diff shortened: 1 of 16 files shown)


Kapin:
Thing to keep in mind, the check for whether to display it or not should validate with the server every single time you navigate to a page that could have it. It should be impossible to reach without superuser status.

Ommy:
@Kapin Realistically if they go into the console and somehow change the user stored in session store, then the backend should still validate the user's status (i.e. in the concept uploader ViewSet so we still get a 401). We could, in the conceptsRouter, create a user model from the user information stored in session store and fetch the model from the backend, and then check on the returned fields.

Ommy:
image

src-backbone/app/js/controllers/conceptsController.js
```diff
@@ -0,0 +1,27 @@
...
+const ConceptsUploaderView = require('views/concepts/conceptsUploaderView');
+
+module.exports = Marionette.Controller.extend({
+
+    routeUploader: function () {
+        if (!App().session.isValid()) {
+            Helpers.navigateToDefaultLoggedOut();
+            return;
+        }
+        App().session.refreshUser(false);
```
Trinovantes: You're freezing the UI thread here. Why not just let the async callback redirect the user instead?

Trinovantes:
Some of the indentation is off

If we had the time, I'd also like to rewrite/restructure how we handle users/session models since they're so non-coherent (e.g. instantiating new user model from session attributes?) but still strongly coupled together right now.

Ommy:
Could move all the functions to User model and put accessor in App()

Kapin:
lgtm. I'd make a ticket to uncouple the user and session objects.

Ommy:
Yeah I'll be doing that, or attempting to do that now

Issue #365, opened 2016-03-04
It should display the user's name or username instead of Trinovantes on the main screen


Kapin:
<img width="134" alt="screen shot 2016-03-04 at 1 04 27 pm" src="https://cloud.githubusercontent.com/assets/995382/13535727/c450877c-e209-11e5-9f50-e2acb78f28e8.png">

PR #371, opened 2016-03-05
Properly get and use the user's name

### LINKED ISSUE #368: Author field should be auto-populated by the user name by default

Kapin:
Change does two key things

1. Adds first name and last name to the signup form
2. Makes the author field on procedure default to owner's full name if it is blank or non-existent.
3. Fixes the "Hello username" field
4. Fixes translations

Closes #368, #365, #380, #381

### DIFF
--- a/src-backbone/app/js/models/procedure.js
+++ b/src-backbone/app/js/models/procedure.js
@@ -12,10 +12,12 @@ let Procedure = Backbone.Model.extend({
     urlRoot: '/api/procedures',
 
     defaults: function() {
+        let user = App().session.user;
+
         // These values are only used when we POST /procedures a new procedure
         return {
             'title': i18n.t('Untitled Procedure'),
-            'author': i18n.t('No author specified'), // TODO either fetch user's username or let API allow nullable authors
+            'author': user.get('first_name') + ' ' + user.get('last_name'),
         };
     },
 
--- a/src-backbone/app/js/views/procedures/proceduresNavbarView.js
+++ b/src-backbone/app/js/views/procedures/proceduresNavbarView.js
@@ -1,4 +1,5 @@
 const EventKeys = require('utils/eventKeys');
+const App = require('utils/sanaAppInstance');
 
 
 module.exports = Marionette.LayoutView.extend({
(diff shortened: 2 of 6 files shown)


Trinovantes:
Either the author field needs to be nullable or the first and last name cannot be null

Kapin:
@Trinovantes that's what I did

Trinovantes:
Then do it on the frontend instead of the backend so the user doesn't initially see an empty text field then a few seconds later get filled when the async request finishes. Ask @Ommy where we store the user name

Kapin:
Need #370 merged first to do that properly.

Ommy:
@Kapin #370 is merged

src-backbone/app/js/models/procedure.js
```diff
@@ -12,10 +12,12 @@ let Procedure = Backbone.Model.extend({
     urlRoot: '/api/procedures',
 
     defaults: function() {
+        let user = App().session.getUser();
```
Ommy: not a method

Kapin: I'm following what you did for the concepts view. I'm confused.

Ommy: https://github.com/SanaMobile/sana.protocol_builder/blob/master/src-backbone/app/js/views/common/loggedInRightNavbarView.js#L10

Kapin: I added that function to session.js. I can do it that way, but I wanted to abstract away the key-value store.

Ommy: Oh oops, my mistake, I didn't see that. Carry on

Kapin: I'll change yours though :smile:

Trinovantes:
is the getUser method still consistent with @Ommy's refactor?

Kapin:
@Trinovantes this should go in before that.

Kapin:
@Trinovantes I'll wait till he merges

Ommy:
@Kapin #387 merged

src-backbone/app/js/views/procedures/proceduresNavbarView.js
```diff
@@ -1,4 +1,5 @@
 const EventKeys = require('utils/eventKeys');
+let App = require('utils/sanaAppInstance');
```
Ommy: const

Kapin: done

Ommy:
+1

PR #376, opened 2016-03-05
Hide the language selector

### LINKED ISSUE #367: Language dropdown should be hidden until we have proper translations

Kapin:
We don't have translations yet... Closes #367


Trinovantes:
Did you even test to see if this works? It's still trying to load the `languageSelectorView` into the `languageSelector` region. 

You only need to comment out line 10 of rightNavbarBehavior.js and comment `fr` from the config file so we can easily add the functionality back later

Kapin:
Closing for now, because I'm dumb.

PR #386, opened 2016-03-06
Add test to prove xml escapes strings

### LINKED ISSUE #293: Sana should escape quotes when generating XML file
```
<Procedure author=""blah" title="test">
</Procedure>
```

Kapin:
Invalidates/Closes #293.

It escapes strings.


Ommy:
+1

Trinovantes:
what about commas in choices and answers fields?

Kapin:
@Trinovantes that is not default behaviour and there is nothing in the spec that states they need to be escaped.

PR #415, opened 2016-03-08
Don't re-render header view when title/author are changed

connorcimowsky:


Kapin:
+1

Trinovantes:
Note that removing this may trigger this race condition (note the title in the list and what's in the input field)

sana

PR #418, opened 2016-03-08
Properly set the value of the required checkbox

### LINKED ISSUE #407: Required checkbox toggle should trigger a save
Came up in user testing, doesn't appear to work properly.

Kapin:
Closes #407


Trinovantes:
Use the handlebars helper `{{ checkSelected selected true }}`

Kapin:
@Trinovantes it needs to return the string "checked" for a checkbox. I can make a helper that does that, but it's different behaviour.

Trinovantes:
+1

PR #423, opened 2016-03-08
Truncate long question text with ellipsis

### LINKED ISSUE #416: Question won't save if the string is too long

Ommy:
image

Resolves #416


Kapin:
+1

Trinovantes:
Isn't the bug caused by the server rejecting long strings?

Kapin:
@Trinovantes we have not been able to reproduce that. There is a copy/paste bug where it is updating twice and sending an empty string the second time.

PR #424, opened 2016-03-08
Added a save button on builder header

### LINKED ISSUE #403: We should add an explicit save changes button even if it doesn't do anything
People expect to press save when done. Came up in user testing. Even if the button doesn't have a meaningful effect.

Ommy:
Save button is right beside the download button, and it just calls the
direct saveToServer function since it isn't autosaving

image

Resolves #403

### DIFF
--- a/src-backbone/app/css/modules/builder.less
+++ b/src-backbone/app/css/modules/builder.less
@@ -71,14 +71,24 @@ header#builder-meta-data{
         }
     }
 
-    a#download-btn{
+    a.builder-header-btn{
         margin-top: -20px;
         position: absolute;
-        top:50%; right: @container-side-padding;
+        top:50%;
 
         font-weight: bold;
         height: 40px;
         line-height: 40px;
         padding: 0 20px;
+    }
+
+    a#download-btn{
+        &:extend(.builder-header-btn);
+        right: @container-side-padding;
+    }
+
+    a#save-btn{
+        &:extend(.builder-header-btn);
+        right: @container-side-padding * 3;
     }
 }
--- a/src-backbone/app/css/modules/builder.less
+++ b/src-backbone/app/css/modules/builder.less
@@ -71,24 +71,25 @@ header#builder-meta-data{
         }
     }
 
-    a.builder-header-btn{
+    div#button-container{
         margin-top: -20px;
         position: absolute;
         top:50%;
-
-        font-weight: bold;
-        height: 40px;
-        line-height: 40px;
-        padding: 0 20px;
-    }
-
-    a#download-btn{
-        &:extend(.builder-header-btn);
         right: @container-side-padding;
+
+        a.btn{
+            float: left;
+            font-weight: bold;
+            height: 40px;
+            line-height: 40px;
+            padding: 0 20px;
+
+            margin-right: @column-margin;
+            &:last-child{
+                margin-right: 0;
+            }
+        }
+
     }
 
-    a#save-btn{
-        &:extend(.builder-header-btn);
-        right: @container-side-padding * 3;
-    }
 }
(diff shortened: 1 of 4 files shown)


Kapin:
Does this only save the procedure parts or do a cascaded save as well?

Ommy:
Oh right, no I have not added that yet. Doing that now

Kapin:
Just one nit (maybe follow-up issue, I can do it). We should have a save element button too on each element view. But this is great. +1

Ommy:
@Kapin would that be a button for each element, if so where exactly would that go. Or are you talking about a save button for each page?

Kapin:
@Ommy let's discuss in person. That should be a completely separate change.

src-backbone/app/css/modules/builder.less
```diff
@@ -71,14 +71,24 @@ header#builder-meta-data{
...
     }
+
+    a#download-btn{
+        &:extend(.builder-header-btn);
+        right: @container-side-padding;
+    }
+
+    a#save-btn{
+        &:extend(.builder-header-btn);
+        right: @container-side-padding * 3;
```
Trinovantes: this is assuming the download button is a fixed width

wrap the buttons with a parent container and absolute position the parent

src-backbone/app/js/views/builder/builderHeaderView.js
```diff
@@ -11,16 +11,27 @@ module.exports = Marionette.ItemView.extend({
...
-        'click @ui.downloadButton': '_download'
+        'click @ui.downloadButton': '_download',
+        'click @ui.saveButton':  '_saveProcedure'
     },
 
     _save: _.debounce(function() { this._saveToServer(); }, Config.INPUT_DELAY_BEFORE_SAVE),
 
+    _saveProcedure: function() {
+        this._saveToServer();
+        this.model.pages.forEach(function(page) {
```
Trinovantes: shouldn't the server just do cascade save instead of us making all these network requests?

src-backbone/app/js/views/builder/builderHeaderView.js
```diff
@@ -11,16 +11,27 @@ module.exports = Marionette.ItemView.extend({
...
         downloadButton: 'a#download-btn',
+        saveButton: 'a#save-btn'
     },
 
     events: {
         'keyup @ui.titleField': '_save',
         'keyup @ui.authorField': '_save',
-        'click @ui.downloadButton': '_download'
+        'click @ui.downloadButton': '_download',
+        'click @ui.saveButton':  '_saveProcedure'
```
Trinovantes: add a trailing comma here

src-backbone/app/css/modules/builder.less
```diff
@@ -71,14 +71,24 @@ header#builder-meta-data{
         }
     }
 
-    a#download-btn{
+    a.builder-header-btn{
```
Trinovantes: just use `.btn` class

Trinovantes:
the page is a thin wrapper so just a save button in each element is enough

src-backbone/app/js/views/builder/builderHeaderView.js
Ommy: I don't understand, the `model.save()` sends the following payload:

```javascript
{
"id":38,
"title":"Untitled Procedure",
"author":"weert grgr",
"uuid":"f6734d7b-b425-4122-a3cf-6dee1020fdbd",
"owner":29,
"last_modified":"2016-03-08T22:05:30.779Z",
"created":"2016-03-08T17:15:01.117Z",
"pages":[218,219,220]
}
```

src-backbone/app/js/views/builder/builderHeaderView.js
Ommy: done

src-backbone/app/css/modules/builder.less
Ommy: done

src-backbone/app/css/modules/builder.less
Ommy: done

Ommy:
+1 - @Trinovantes

PR #425, opened 2016-03-08
Explicitly make all foreign key deletes cascade

Kapin:
It defaults to this, but better for my piece of mind and it becomes mandatory to do this in Django 2.0


Ommy:
lgtm for the most part. Can we add some tests for this functionality?

Kapin:
Yes

Trinovantes:
should also remove references of elements from conditions

Kapin:
@Trinovantes that is much more complicated as there isn't really a way of easily determining whether an element has references to conditions right now.

PR #430, opened 2016-03-08
Trigger render on left-nav when an element is updated

Kapin:
Fixes the issue where updating a question on an element doesn't actually update the left-nav.


Ommy:
+2

PR #435, opened 2016-03-09
Remove eid and just use the element's primary key instead

### LINKED ISSUE #404: Element id should just be handled behind the scenes
We can just use the pk of the element. We don't need to generate anything special.

### LINKED ISSUE #416: Question won't save if the string is too long

Kapin:
Simplifies logic.

Closes #416 
Closes #404


Ommy:
+2

PR #446, opened 2016-03-09
Re-render the header once, to populate the title and author

### LINKED ISSUE #444: Title of Procedure is not updating properly when returning to it

Kapin:
Closes #444

### DIFF
--- a/src-backbone/app/js/views/builder/builderHeaderView.js
+++ b/src-backbone/app/js/views/builder/builderHeaderView.js
@@ -21,7 +21,14 @@ module.exports = Marionette.ItemView.extend({
         'click @ui.saveButton':  '_saveProcedure',
     },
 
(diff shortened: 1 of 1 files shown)


src-backbone/app/js/views/builder/builderHeaderView.js
```diff
@@ -32,6 +37,8 @@ module.exports = Marionette.ItemView.extend({
         });
     },
 
+    _renderOnce: _.once(function() { this.render(); }),
```
connorcimowsky: Would prefer the function parameter to be on multiple lines.

connorcimowsky:
LGTM once @Trinovantes weighs in on whether this will fix the race condition once and for all

src-backbone/app/js/views/builder/builderHeaderView.js
Kapin: can do

Trinovantes:
probably +1

PR #452, opened 2016-03-09
Cap i18next version

### LINKED ISSUE #450: TypeError: i18n.use is not a function
Build is failing. No idea why.

```
 1)  "before each" hook for "should redirect user to logged in page if there is a token":
     TypeError: i18n.use is not a function
      at Object._setupI18Next (setup.js:74:14)
      at Object.loadLibs (setup.js:24:14)
      at Context.<anonymous> (hooks.js:19:28)
```

Trinovantes:
Closes #450


Ommy:
:shipit:

PR #457, opened 2016-03-11
Add a check for empty responses in parse functions

connorcimowsky:
For some reason, only in Safari, `parse` is called with an empty response and then subsequently called with the populated response. This caused errors to be thrown, so I added a check for an early return.

EDIT: Looks like this is more complicated, let's hold off for now.


Ommy:
+1

Trinovantes:
What errors are thrown? And what does the network console say?

connorcimowsky:
"LGTM" —@Trinovantes

PR #458, opened 2016-03-11
Make popovers bigger

connorcimowsky:
Can't seem to make them size to fit, but this will at least make them fit most conditionals. ¯\\_(ツ)_/¯


Kapin:
+1

PR #464, opened 2018-01-05
Add unique constraint to concept name

xjie11:
Added unique constraint to concept name
Added script to change name and display name for duplicates
Added migration


xjie11:
wrong branch

